Issue #3, opened 2019-10-19
Should we use typescript ?


romainPrignon:
- Transition to Deno will be easier

ry:
Yeah - eventually. I'm currently in prototype mode. I'd rather move fast and break things for now.

PR #4, opened 2019-10-19
[WIP] chore: allow and lint typescript files

romainPrignon:
- add typescript configuration based on the react one
- use the `typescript-eslint` parser for eslint

---
When building the react app with `npm run build`, react will automatically rewrite the content of `tsconfig.json` if it don't like it.

Here is the command output:
```
> react-scripts build

The following changes are being made to your tsconfig.json file:
  - compilerOptions.allowJs to be suggested value: true (this can be changed)
  - compilerOptions.skipLibCheck to be suggested value: true (this can be changed)
  - compilerOptions.allowSyntheticDefaultImports to be suggested value: true (this can be changed)
  - compilerOptions.forceConsistentCasingInFileNames to be suggested value: true (this can be changed)
  - compilerOptions.module must be esnext (for import() and import/export)
  - compilerOptions.resolveJsonModule must be true (to match webpack loader)
  - compilerOptions.isolatedModules must be true (implementation limitation)
  - compilerOptions.noEmit must be true
  - compilerOptions.jsx to be suggested value: react (this can be changed)
```

I use the suggested value from react. We May need to twink it

---

ref: To be merged when ready to switch to TS (see #3)

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,17 @@
+{
+  "parser": "@typescript-eslint/parser",
+  "parserOptions": {
+    "ecmaVersion": 2019,
+    "sourceType": "module",
+    "project": "./tsconfig.json",
+    "ecmaFeatures": {
+      "jsx": true
+    }
+  },
+  "plugins": [
+    "@typescript-eslint"
+  ],
+  "extends": [
+    "react-app"
+  ]
+}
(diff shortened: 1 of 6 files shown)


ry:
Thanks - but I prefer to be less strict during this prototyping phase

ry:
Thanks but I ended up needing ts a bit earlier and so these changes were largely done in https://github.com/denoland/deno_website2/commit/5b6956837153614dd7d060707aa76c03b9c42653

PR #7, opened 2019-10-19
Fix pre tag style

themzed:
- Add style to pre tag
- Fix nested pre tags style on SyntaxHighlighter


ry:
Thanks - fixed the unnecessary pre tag in https://github.com/denoland/deno_website2/pull/10

I'm trying to keep the css as minimal as possible until we can have a designer look into it.

Issue #26, opened 2019-10-22
bring back "normalize" feature in the benchmark plots 


ry:
as was done in https://github.com/denoland/deno/pull/2919

cc @lucacasonato

lucacasonato:
I'll take this on - should be a lot cleaner in React.

lucacasonato:
So there are some issues with c3-react. The C3Chart component does not react to state updates (like the label) - this means that when I change for example the label, the chart does not change. I can also not use multiple charts and just hide/unhide them like I did on the original site because then c3 does not properly size the chart and content overflows on the bottom. Do you have any ideas on how we could deal with this? We should consider using a different chart library that is more reactive if we can't figure this with c3.

ry:
@lucacasonato That sucks. I don't have any ideas. I'm happy to explore a different charting module.

lucacasonato:
I'll look into some other chart modules then. I saw some interesting ones while researching this issue yesterday.

lucacasonato:
Fixed by #39

PR #39, opened 2019-10-24
Moved to new chart library and readded normalized charts

lucacasonato:
Changed the chart library to  [apexcharts](https://apexcharts.com) - this is a pretty complete library and works way better with React than c3 (discussed in #26).

This adds some nice things: 
- we now have png and svg download buttons on the charts (provided by the apexchart module out of the box)
- way more customization options https://apexcharts.com/docs/options
- better zooming than c3 - only supported on numeric x-axis so we aren't using it right now - I am planning to re add this in a future PR.
- animations if we want them. They seem to slow everything down quite a bit, so I disabled them for now - we can look into that later as well.

I also re added the normalized checkbox and the respective graphs.


lucacasonato:
The build is failing because of issue described in #41

ry:
@lucacasonato please merge master

ry:
@lucacasonato Looks fresh. Is this ready to land?

lucacasonato:
@ry Sure

PR #41, opened 2019-10-24
Only try to publish to staging when email and key are present

lucacasonato:
Currently CI fails for users that are not part of the denoland github group. This fixes that by only trying to publish to cloudflare if email and cloudflareKey are present.


lucacasonato:
@ry What I did here doesn't work as secrets can't be accessed at all from external PR - this means that there is no good way to check if secrets are going to be available or not. Deploy staging would pretty much only work for you and Bert - how about removing the stage all together and only running staging deploys locally?

PR #46, opened 2019-10-24
Fix CI for other users

ry:
ref https://github.com/denoland/deno_website2/pull/41


ry:
@lucacasonato does this solve your problem?

PR #51, opened 2019-10-25
Add Linux support on Homebrew

juanarbol:
Just as [brew.sh](brew.sh) says: "The missing package manager for macOS (or Linux)"


ry:
How about just remove the parens (mac) instead

juanarbol:
Jm, sounds better.

juanarbol:
Ah no, we should leave Mac or Linux because It only works on Linux or macOS

PR #54, opened 2019-10-25
Add basic docs sidebar

kevinkassimo:
Add basic closable sidebar for navigation.

deno_web_1

Also introduced `node-sass` to make writing styles easier.

PR #56, opened 2019-10-25
bring back kevin's sidebar patch

ry:
can't figure out the margins.

rebased from #54 
cc @kevinkassimo

### DIFF
--- a/src/Docs.tsx
+++ b/src/Docs.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React from "react";
 import { main } from "./doc_utils";
 import CodeBlock from "./CodeBlock";
 import Markdown from "./Markdown";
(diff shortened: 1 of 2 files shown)


src/Docs.tsx
```diff
@@ -22,19 +25,20 @@ interface Props {
 export default function Docs(props: Props) {
   const location = useLocation();
   const docs = main(location.pathname, props.source);
+  const [drawerOpen, setDrawerOpen] = useState(true);
```
kevinkassimo: We don’t need this if we don’t ever want to hide the drawer

kevinkassimo:
I believe the layout is strange potentially because of the Container maxWidth md at the top level. I did not change it in the original PR since it involves greater subjective structure changes

ry:
closed in favor of #100

PR #60, opened 2019-10-25
feat: add dark mode

lucacasonato:
I have re-added dark mode. It gets enabled automatically by the media query `prefers-color-scheme: dark`. It was relatively easy to add after moving all <a> tags to <Link> tags from material-ui. Now material styling also gets applied to links.

The only CSS I added was to invert the logo. Both the code blocks and the graphs have built in dark modes, so those were relatively easy - they get triggered by `prefers-color-scheme: dark` too.


ry:
Not sure what happened but I think this commit broke the links in https://deno.land/std/

lucacasonato:
@ry It did. I'll submit a fix in a second.

bartlomieju:
It's cool that dark mode is back but I liked the old one much better. White text on black background is really tiring for eyes. Any chance we could change background color to something more toned down? Even using color from charts for background color looks better IMHO

Zrzut ekranu 2019-10-25 o 22 45 30

Zrzut ekranu 2019-10-25 o 22 44 48

vladimyr:
I second what @bartlomieju already said plus color contrast of syntax highlighting is really low. That is why I tried to switch to dark mode and how I ended up here.
Having dark mode that requires me to change my operating system UI settings to activate it or install a  browser extension that would somehow trick site is basically like not having it at all: https://web.dev/prefers-color-scheme/#dark-mode-but-add-an-opt-out

PR #62, opened 2019-10-25
Fix links in registry folder view

lucacasonato:
Fixes issue introduced in #60


ry:
Thanks for the quick fix! 

This is the sort of thing I hope we can protect against with tests

ry:
@lucacasonato I think there's another link issue with the breadcrumbs too. Try clicking on "std" from https://deno.land/std/bundle/

lucacasonato:
Just found that too. I'll submit a fix in a few minutes.

PR #80, opened 2019-10-28
Changed db path entries to have no leading slash

### LINKED ISSUE #79: 404 errors from date_fns JS files in registry
If you visit this page: https://deno.land/x/date_fns/addMonths/index.js you'll see a 404 error.

This reporter seemed to think it was an issue with JS files: https://github.com/denoland/registry/issues/177

However, if you look at the "repository" link on the `index.js` page linked above, you'll see it points to: https://github.com/date-fns/date-fns/tree/mastersrc/addMonths/index.js

Note `/tree/mastersrc/`, which should actually be `/tree/master/src/`. I suspect this might have something to do with the handling of the `path` parameter in database.json: https://github.com/denoland/deno_website2/blob/92a20e81c51c051080311c2edefe517049dbafcd/src/database.json#L116

Note the old registry code has a leading `/`: https://github.com/denoland/registry/blob/714ffcf29d55620fb02b53c99c7d954a366a19e7/src/database.json#L116

lucacasonato:
Changed registry to expect path entries to have no leading slash. Also changed `std` entry in the database to have no leading slash.

Fixes #79

### DIFF
--- a/src/database.test.ts
+++ b/src/database.test.ts
@@ -14,7 +14,7 @@ test("a database entry should never have a path starting with /", () => {
   for (const key in DATABASE) {
     const entry = DATABASE[key];
     if (entry.path) {
-      expect(entry.path[0]).not.toMatch("/");
+      expect(entry.path[0]).not.toBe("/");
     }
   }
 });
--- a/src/database.test.ts
+++ b/src/database.test.ts
@@ -1,0 +1,20 @@
+import DATABASE from "./database.json";
+
+test("a database entry of type github should have a owner and repo", () => {
+  for (const key in DATABASE) {
+    const entry = DATABASE[key];
+    if (entry.type == "github") {
+      expect(entry.owner).toBeTruthy();
+      expect(entry.repo).toBeTruthy();
+    }
+  }
+});
+
+test("a database entry should never have a path starting with /", () => {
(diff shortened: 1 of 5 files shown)


ry:
Thanks - Would you mind adding a test that ensures future conformance when people add new database entries ?

src/database.test.ts
```diff
@@ -0,0 +1,20 @@
...
+      expect(entry.repo).toBeTruthy();
+    }
+  }
+});
+
+test("a database entry should never have a path starting with /", () => {
+  for (const key in DATABASE) {
+    const entry = DATABASE[key];
+    if (entry.path) {
+      expect(entry.path).not.toMatch(/^\/.*/);
```
ry: Nit: it seems more readable to just check the first character than use a regular expression

lucacasonato: I agree - Jest does not have that built in though. I could rewrite it as `expect("entry.path".startsWith("/")).toBeFalsy();` or we could get this matcher and a lot more matcher by using the [jest-extended](https://github.com/jest-community/jest-extended) library. What do you prefer?

ry: expect(entry.path[0]).toBe("/")

lucacasonato: That works too :-)

PR #91, opened 2019-10-29
Fixing link component 

### LINKED ISSUE #86: Link to master branch should not be internal
Clicking the "master branch" link in the Benchmarks page treats the link as a route: `https://deno.land/https://github.com/denoland/deno`.

I think this should be an `href` instead.
https://github.com/denoland/deno_website2/blob/d0b37acea76c2f80866d99a4d89859b7bc26981a/src/Benchmarks.tsx#L124

josephsintum:
#86 

Making a custom link component that determines internally if it is a internal or external link and automatically uses the correct tag and correct styling and avoid future errors

`touch /src/Link.tsx`

```tsx
import React from "react";
import { HashLink, HashLinkProps } from "react-router-hash-link";
import { Link as MaterialLink } from "@material-ui/core";

const InternalLink = React.forwardRef((props: HashLinkProps, ref) => (
  <HashLink {...props} />
));

export const Link = (props) => {
  if (props.to.indexOf("://")=== -1) {
    return <MaterialLink component={InternalLink} {...props}/>
  } else {
    return <MaterialLink {...props} component="a" href={props.to} />
  }
}
```

then usage is really simple

```tsx
import React from "react";
import { Link } from "./Links";

export const RandomComponent = (props) => {
  <>
      <Link to="/benchmarks">Styled Link</Link>
      <Link to="https://github.com/denoland/deno">master branch</Link>
  </>
}```

### CODE REVIEW

**Comment by lucacasonato** [comment]
Looks good!

What do you think about also adding a `<Button>` component that uses MaterialButton? That way we get the same improvements for buttons too, not just links.

Oh and it would be cool if you could also add some tests :-).

File: src/Link.tsx [hunk 1]
```diff
@@ -0,0 +1,15 @@
+import React from "react";
+import { HashLink, HashLinkProps } from "react-router-hash-link";
+import { Link as MaterialLink } from "@material-ui/core";
+
+const InternalLink = React.forwardRef((props: HashLinkProps, ref) => (
+  <HashLink {...props} />
+));
+
+export const Link = (props) => {
+  if (props.to.indexOf("://")=== -1) {
```
ry: Can you run "yarn format"?

josephsintum: sure

ry:
can you also update the links elsewhere? For example there are many in src/Home

josephsintum:
> Looks good!
> 
> What do you think about also adding a `<Button>` component that uses MaterialButton? That way we get the same improvements for buttons too, not just links.
> 
> Oh and it would be cool if you could also add some tests :-).



> Looks good!
> 
> What do you think about also adding a `<Button>` component that uses MaterialButton? That way we get the same improvements for buttons too, not just links.
> 
> Oh and it would be cool if you could also add some tests :-).

I'll add <Button> to it

josephsintum:
I realized a lot people are using "href" or "to"
```tsx
<Link to="somewhere">some where</Link>
or 
<Link href="somewhere">some where</Link>
```

Should I enforce consistency by allow only <Link to=""> 
or allow flexibility with both

ry:
> Should I enforce consistency by allow only or allow flexibility with both

Consistency would be preferable. Maybe you can do that by defining an interface for the props.

josephsintum:
> > Should I enforce consistency by allow only or allow flexibility with both
> 
> Consistency would be preferable. Maybe you can do that by defining an interface for the props.

I have added the interface and added check to prevent "href" use

src/components/Link.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+export const InternalLink = React.forwardRef((props: HashLinkProps, ref) => (
+  <HashLink {...props} />
+));
+
+interface Props {
+  /** Supports both internal and external links */
+  to: string;
+  /** Do not use href. Use <Link to="link" > instead for both internal and external links */
+  href?: never;
```
ry: You can just leave this out.

src/components/Link.tsx
```diff
@@ -0,0 +1,23 @@
+import React from "react";
```
ry: Do not introduce a new directory - if you want to make a new organization it should be done in a separate PR.

src/components/Link.tsx
```diff
@@ -0,0 +1,23 @@
...
+));
+
+interface Props {
+  /** Supports both internal and external links */
+  to: string;
+  /** Do not use href. Use <Link to="link" > instead for both internal and external links */
+  href?: never;
+}
+
+export const Link: React.FC<Props & LinkProps> = props => {
```
ry: Maybe add some documentation here saying

```
/** Use this component instead of MaterialUI's Link or ReactRouter's Link. */
```

src/components/Link.tsx
josephsintum: taking it out

src/Button.tsx
```diff
@@ -0,0 +1,21 @@
+import React from "react";
+import { InternalLink } from "./Link";
+import { Button as MaterialButton } from "@material-ui/core";
+import { ButtonProps } from "@material-ui/core/Button/Button";
+
+interface Props {
+  /** Supports both internal and external links */
+  to?: string;
+  component?: any;
```
ry: s/any/React.Component/

josephsintum: component?: HTMLElement;

src/Button.tsx
```diff
@@ -0,0 +1,22 @@
...
+}
+
+/** Use this component instead of MaterialUI's Button or ReactRouter's Link */
+export const Button: React.FC<Props & ButtonProps> = props => {
```
lucacasonato: @ry What do you think of this? This just makes sure you don't use href by accident.
```suggestion
export const Button: React.FC<Props & Omit<ButtonProps, 'href'> = props => {
```
If this is changed make sure to change it [here](https://github.com/denoland/deno_website2/pull/91/files#diff-01e857eda18d514416246bb1ff1e7510R16) too.

ry: Why can't we just use `<a href>`... I don't understand why getting a link is requiring so much infrastructure.

lucacasonato: @ry https://github.com/denoland/deno_website2/issues/86#issuecomment-547166068

ry: Thank you @lucacasonato - I missed that when you originally posted it. That was well put and very much helps me justify this.


@josephsintum - could you please copy @lucacasonato's comment into the top of src/Link.ts ? (since it is not at all obvious why this is necessary)

ry: I don't understand why react-router's Link doesn't understand the difference between external and internal links tho.

src/Button.tsx
```diff
@@ -0,0 +1,22 @@
...
+  to?: string;
+  component?: HTMLElement;
+}
+
+/** Use this component instead of MaterialUI's Button or ReactRouter's Link */
+export const Button: React.FC<Props & Omit<ButtonProps, 'href'>> = props => {
+  let { to, ...rest } = props;
+  if (!props.to) {
+    return <MaterialButton {...props} />;
+  } else if (props.to.indexOf("://") === -1) {
```
ry: this check should only be in one place in our code base.

src/Button.tsx
josephsintum: React doesn't handle routing 
So if `<a href>` is used, react will render the anchor tag without the SPA routing logic, then the whole page is reloaded which is not a great experience
so we use external packages like react-router which provides a `<Link to="...">` to handle routing, 
The link need to be styled by material ui, so react-router <Link> is wrapped with material ui <Link>

lucacasonato: > I don't understand why react-router's Link doesn't understand the difference between external and internal links tho.

I agree - that is stupid, they should just handle external links as external links.

src/Button.tsx
lucacasonato: @josephsintum HTMLElement is a default HTML element not a React component - you should use `React.ComponentType<any>`

ry:
LGTM - thanks @josephsintum and @lucacasonato

PR #92, opened 2019-10-29
refactor: use recharts for benchamark charts

### LINKED ISSUE #61: Benchmark charts - bad values for throughput and exec time
Zrzut ekranu 2019-10-25 o 13 25 13

If you check latest benchmarks (https://github.com/denoland/deno/runs/274208380) you can see that both of those charts should be <2s but all of the results are >4s on charts.

CC @lucacasonato @ry

### LINKED ISSUE #84: benchmark plots don't display well on mobile
they're not using all the available horizontal real-estate 

Screen Shot 2019-10-28 at 1 23 57 PM

lucacasonato:
`recharts` is a well maintained and reasonably simple react wrapper for d3.

I know we moved to apexcharts only a few days ago, but we should go back to a d3 based library as it is way more stable. This should be the last chart library move we will have to make as d3 is stable and very extensible so if something isn't right we can just change it. That was not the case with `apexcharts` - it is very limiting in what you can actually customize and where you can hook into the render.

I re implemented all of the `apexcharts` features we had:
- hover label to highlight a specific series
- disable certain series by clicking on them

I also added zooming again which we didn't have with `apexcharts`. This also fixes #61.

Re-implementing the download svg and download png buttons from `apexcharts` with `recharts` is trivial, but I just didn't do it yet - I can do that in a later PR.


ry:
Switching chart modules sounds good - but the mobile margin issues seem to not be fixed 
Screen Shot 2019-10-29 at 11 34 20 AM

ry:
I'm removing "fixes #84" from the description so this can be landed

lucacasonato:
> Just saw that scrolling on mobile is kinda annoying because of the chart zooming. If you touch on the chart while scrolling then you either trigger a zoom or a link to github - less than ideal. I need to fix that before this can be merged.

Should I try fix this first, or do that in a separate PR?

lucacasonato:
@ry Should we even have zooming and the popup on small screens? The popup just blocks the view of the entire chart because it is so big, and trying to zoom accurately on a small screen is close to impossible. I can't think of any good way to differentiate a scroll motion and a zoom motion on touch devices. They are so similar.

ry:
@lucacasonato re zoom popup, I don't find it helpful...

> Should I try fix this first, or do that in a separate PR?

Might as well do it in this one - if we're on the same page. Basically the y-axis label takes up too much space on mobile it's almost 1/3 the horizontal width. We used to have it working better...

src/Benchmarks.tsx
```diff
@@ -21,87 +28,173 @@ interface Props {
...
-      }
-    },
-    xaxis: {
-      labels: {
-        show: false
-      },
-      categories: props.sha1List,
-      tooltip: {
-        enabled: false
+  function zoom() {
```
ry: Isn't there a library that implements zoom for us? I don't want to maintain UI code.

src/Benchmarks.tsx
```diff
@@ -21,87 +28,173 @@ interface Props {
...
+            domain={["auto", "auto"]}
+            stroke={theme.palette.getContrastText(
+              theme.palette.background.default
+            )}
+          />
+          {refLeft !== null && refRight !== null ? (
+            <ReferenceArea x1={refLeft} x2={refRight} />
+          ) : null}
+        </LineChart>
+      </ResponsiveContainer>
```
ry: This looks overly complicated...

ry:
closing this because it's old. @lucacasonato please reopen if you want to get it landed.

lucacasonato:
I am so sorry I wasn't able to get back to this earlier. I'll try to figure out the issue with the excessive left spacing on the mobile view and then I'll repopen the PR. I am hoping to get to this asap, but im relativly busy atm.

PR #99, opened 2019-11-01
Cache dependencies in GitHub Actions

imbsky:


ry:
Hi - I don't know what this does. Can you explain?

imbsky:
Use the node_modules cache and skip the package install step if the hash value of yarn.lock does not change.

ry:
I don't want to cache things now because it creates additional complexities - the build doesn't take very long. But thanks for the patch - I didn't know about this feature. I think we will use it in the future in deno or [rusty_v8](https://github.com/denoland/rusty_v8).

PR #125, opened 2019-12-01
allow multiple declarations

zhmushan:
[this link is broken](https://deno.land/x/abc/http_exception.ts?doc)
multiple declarations mean functions are overloaded

### DIFF
--- a/src/doc.test.ts
+++ b/src/doc.test.ts
@@ -54,3 +54,97 @@ test("enum", () => {
     }
   ]);
 });
+
+test("multiple declarations", () => {
+  const rootModule = "http_exception.ts";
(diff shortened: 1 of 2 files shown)


ry:
What does this fix? Can you add a test?

zhmushan:
There are function overloads in this file, so a function has multiple declarations.  But the original code only allowed each function to be declared once. I’m not sure how to test it

ry:
@zhmushan Here is an example: https://github.com/denoland/deno_website2/blob/591b49b8de66c0f87f8fe48195978c76f87c88dd/src/doc.test.ts#L5-L19

PR #127, opened 2019-12-02
Add `noble` cryptographic libraries.

paulmillr:
Hey there!

Adding a few awesome modules. The basic idea was that indunty/elliptic and friends have like 100 dependencies each, and every dep there could get hacked and screwed. One thing that's extremely important for cryptographic primitives is trustworthiness, and modern libs don't have it.

So, i've created [`noble`](https://github.com/paulmillr/noble-ed25519) libraries - they don't use any deps, are self-contained and written in TypeScript. All of them use built-in js bigints, so this is something not yet supported by almost any competition lib. Another goal for noble was to write code that can be audited by non-cryptographers. Easy!

### DIFF
--- a/src/database.json
+++ b/src/database.json
@@ -29,6 +29,11 @@
     "owner": "chiefbiiko",
     "repo": "blake2b"
   },
+  "bls12-381": {
+    "type: "github",
+    "owner": "paulmillr",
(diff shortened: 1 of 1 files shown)


paulmillr:
Thanks Ryan!

PR #134, opened 2019-12-07
sort database.json

justjavac:


ry:
While you're at it, maybe you can add a test to ensure it's always sorted in the future?

justjavac:
done

PR #137, opened 2019-12-08
module name must not contain dashes

justjavac:
cc @paulmillr

### DIFF
--- a/src/database.json
+++ b/src/database.json
@@ -29,7 +29,7 @@
     "owner": "chiefbiiko",
     "repo": "blake2b"
   },
-  "bls12-381": {
+  "bls12_381": {
     "type": "github",
     "owner": "paulmillr",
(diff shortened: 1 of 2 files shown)


paulmillr:
Thanks for the correction!

rsp:
Is the idea to use different names in deno.land/x than on GitHub when the repo names contain dashes?

Wouldn't it be confusing when the same modules are used directly from GitHub or using services like GitHub Packages, pika.dev or other package services in the future? None of which automatically change the package name to my knowledge. I know that I could publish a module on deno.land/x with a completely different name than the name of the repo, just like I can on npm, but I don't really understand the reason to exclude the most popular character.

Note that dashes are actually the most popular characters in module names on npm as I showed on denoland/deno#1170 so we can expect them to be used more frequently when there are more Deno modules published.

PR #141, opened 2019-12-12
Add SQLite wrapper for Deno

dyedgreen:
Add library for running SQLite from Deno: https://github.com/dyedgreen/deno-sqlite

### DIFF
--- a/src/database.json
+++ b/src/database.json
@@ -494,6 +494,11 @@
     "owner": "manyuanrong",
     "repo": "sql-builder"
   },
+  "sqlite": {
+    "type": "github",
+    "owner": "dyedgreen",
+    "repo": "deno-sqlite"
+  },
   "std": {
     "type": "github",
     "owner": "denoland",


bartlomieju:
@dyedgreen I just tested it and it is awesome! 

May I suggest providing additional TS typings for API so it can be used more easily in editors?

dyedgreen:
Thanks for the quick replay @bartlomieju! I'm glad you like it 😀

Yeah, the plan is to eventually move the whole thing over to TypeScript, but since I'm new to TypeScript I haven't done that yet. (Especially since I imagine proper typing for the `Rows` iterator might be a bit more tricky.)

I moved the `mod.js` to `mod.ts`. This way I can gradually upgrade the internals, but users of the library don't have to know about it and can just `import {open} from "./mod.ts"` from day one! (Moving the other files require some non-trivial changes and I need to read a bit more into TypeScript before doing that 😅)

PR #142, opened 2019-12-13
Repair the Unexpected entry and provide error fallback.

alexjeffburke:
The Unexpected entry was not being rendered because the type of the
entry is "esm" not "github" and we provide no valid repo. Fix the
repo/owner properties in the database but, since there are no useful
files in the repo wire up displaying README.md in our index page.

While here I also discovered that any missing type would actually
cause an exception because it fell off the end of renderDir(). This
case was actually covered in the older registry repo, so port that.

### DIFF
--- a/src/Registry.js
+++ b/src/Registry.js
@@ -67,8 +67,9 @@ export default function Registry() {
   if (isLoading) {
     contentComponent = <Spinner />;
   } else if (state.dir) {
+    const { body, files } = state.dir;
     const entries = [];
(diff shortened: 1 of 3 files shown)


alexjeffburke:
@ry awesome, thanks!

hayd:
Is this broken again? I am not really sure how esm is supposed to work (as part of the db/registry)...

ry:
@hayd What's broken? Can you make an issue?

PR #145, opened 2019-12-16
code split and optimization first-page load

### LINKED ISSUE #144: Page loads too slowly
The current page load resource is too slow

And the loaded resource pack is too large

截屏2019-12-1700 30 42

we should add `babel-import-plugin` plugin

Load components on demand

Then the `component`/`router` should be loaded dynamically

eg. `Chart component`

axetroy:
close #144

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,13 +1,28 @@
 import React from "react";
-import Home from "./Home";
-import Registry from "./Registry";
-import RegistryIndex from "./RegistryIndex";
 import useHashLink from "./useHashLink";
-import NotFound from "./NotFound";
 import PathBreadcrumbs from "./PathBreadcrumbs";
-import Benchmarks from "./Benchmarks";
 import { Container } from "@material-ui/core";
 import { BrowserRouter, Redirect, Switch, Route } from "react-router-dom";
+import Spinner from "./Spinner";
+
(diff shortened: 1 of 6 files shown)


config-overrides.js
```diff
@@ -0,0 +1,16 @@
...
+module.exports = override(
+  fixBabelImports("@material-ui/core", {
+    libraryName: "@material-ui/core",
+    // Use "'libraryDirectory': ''," if your bundler does not support ES modules
+    libraryDirectory: "esm",
+    camel2DashComponentName: false
+  }),
+  fixBabelImports("@material-ui/icons", {
+    libraryName: "@material-ui/icons",
+    // Use "'libraryDirectory': ''," if your bundler does not support ES modules
```
ry: I’m not sure what this comment means. Can it be removed?

package.json
```diff
@@ -6,8 +6,8 @@
     "publish_prod": "wrangler publish --env=production",
     "publish_staging": "wrangler publish --env=staging",
     "publish": "wrangler publish",
-    "start": "react-scripts start",
-    "build": "react-scripts build",
+    "start": "react-app-rewired start",
+    "build": "react-app-rewired build",
```
ry: What does this provide ?

PR #150, opened 2019-12-22
update the database with new library soxa for deno

fakoua:
https://github.com/fakoua/soxa/blob/master/README.md

### DIFF
--- a/src/database.json
+++ b/src/database.json
@@ -509,6 +509,11 @@
     "owner": "manyuanrong",
     "repo": "deno-smtp"
   },
+  "soxa": {
+    "type": "github",
+    "owner": "fakoua",
+    "repo": "soxa"
+  },
   "sql_builder": {
     "type": "github",
     "owner": "manyuanrong",


ry:
@axetroy can you explain?

axetroy:
> @axetroy can you explain?

This is not a Deno module

- Not written in Typescript
- No type declaration
- No test cases

It only works on node.js

So I don't think it is suitable for adding to database.json at this time.

fakoua:
Thank you very much,
I will try my best to convert it the code to typescript, it works well with deno, it is a clone of axios with fetch adapter instead of http module.

However, it is missing test cases and typescript support. 

Thanks again.

ry:
@fakoua TS is not required but some tests in deno (maybe even CI) would be good

see this for an example of how to setup ci https://github.com/oakserver/oak/blob/master/.travis.yml

fakoua:
Thank you very much,
One more question, how can I use deno.land/x/.... Before my pack get listed?

axetroy:
> Thank you very much,
> One more question, how can I use deno.land/x/.... Before my pack get listed?

You don't need to import from `deno.land/x/`

You can import from relative path

fakoua:
Thank you all

fakoua:
Hi,
1- I added test cases for most of modules
2- Automated build/test with travis-ci
3- Test the functionality with DENO and removed any node dependencies.
4- Update the README to include more working examples.

Many thanks

PR #156, opened 2019-12-31
Add 'dargs' Argument parser for Deno

fakoua:
### Example:

```bash
$ deno myfile.ts -x 3 -y 4 -n5 -abc --beep=boop foo bar baz
{ _: [ 'foo', 'bar', 'baz' ],
  x: 3,
  y: 4,
  n: 5,
  a: true,
  b: true,
  c: true,
  beep: 'boop' }
```

### DIFF
--- a/src/database.json
+++ b/src/database.json
@@ -124,6 +124,11 @@
     "owner": "chiefbiiko",
     "repo": "curve25519"
   },
+  "dargs": {
+    "type": "github",
+    "owner": "fakoua",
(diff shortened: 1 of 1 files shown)


bartlomieju:
It's almost the same as [`flags` from standard library](https://github.com/denoland/deno/tree/master/std/flags)...

fakoua:
May thanks for your comment, 
you are right, will close the pull.

PR #166, opened 2020-01-05
Fix the link issue

fakoua:
With the newly introduced link component we need to use "to" instead of "href"
I fixed the issue.

### DIFF
--- a/src/database.json
+++ b/src/database.json
@@ -150,17 +150,6 @@
     "path": "deno/",
     "desc": ""
   },
-  "cowsay": {
-    "type": "github",
-    "owner": "fakoua",
-    "repo": "cowsay"
-  },
-  "cursornext": {
-    "type": "github",
-    "owner": "clitetailor",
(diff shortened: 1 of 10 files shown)


src/database.json
```diff
@@ -150,17 +150,6 @@
...
-    "type": "github",
-    "owner": "fakoua",
-    "repo": "cowsay"
-  },
-  "cursornext": {
-    "type": "github",
-    "owner": "clitetailor",
-    "repo": "cursornext",
-    "path": "deno/"
-  },
```
ry: I think these shouldn't be removed?

src/database.json
```diff
@@ -359,11 +343,6 @@
     "repo": "deno_free_port",
     "desc": "get free port for Deno"
   },
-  "free_port": {
-    "type": "github",
-    "owner": "axetroy",
-    "repo": "deno_free_port"
-  },
```
ry: shouldn't be removed

PR #177, opened 2020-01-07
Simple registry search

### LINKED ISSUE #176: wanted: Add search input to registry page
We already have more than 120 modules listen in registry. It'd be very useful to have a small search box at the top of the registry page than you could filter out the list.

I'd imagine that search box is focused by default when you visit the page and can be used to filter by name/description fields of the module.

lucacasonato:
Added simple registry search with autofocus as requested in #176. Searches name and description exactly. If we want more coarse searching we can use https://fusejs.io/.

Fixes #176.

### DIFF
--- a/src/page/RegistryIndex.js
+++ b/src/page/RegistryIndex.js
@@ -9,7 +9,11 @@ import TextField from "@material-ui/core/TextField";
 
 export default function RegistryIndex() {
   const [query, setQuery] = React.useState("");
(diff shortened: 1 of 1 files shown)


fakoua:
I think the autofocus is annoying on mobile devices since it open the keyboard automatically and hide the page. Consider removing the autofocus flag.
@lucacasonato

PR #180, opened 2020-01-09
feat: Add /api/versions endpoint to worker that returns recent Deno versions 

### LINKED ISSUE #138: Wanted: provide a list of version information
related: https://github.com/denolib/setup-deno/issues/11
Using Github API will limit the rate.

lucacasonato:
This is an idea for how we could do #138.

You have to have 'Accept: application/json' in your headers. Returns the 10 most recent versions like this:

```json
[{"tag":"v0.29.0","url":"https://github.com/denoland/deno/releases/tag/v0.29.0"},{"tag":"v0.28.1","url":"https://github.com/denoland/deno/releases/tag/v0.28.1"},{"tag":"v0.28.0","url":"https://github.com/denoland/deno/releases/tag/v0.28.0"},{"tag":"v0.27.0","url":"https://github.com/denoland/deno/releases/tag/v0.27.0"},{"tag":"v0.26.0","url":"https://github.com/denoland/deno/releases/tag/v0.26.0"},{"tag":"v0.25.0","url":"https://github.com/denoland/deno/releases/tag/v0.25.0"},{"tag":"v0.24.0","url":"https://github.com/denoland/deno/releases/tag/v0.24.0"},{"tag":"v0.23.0","url":"https://github.com/denoland/deno/releases/tag/v0.23.0"},{"tag":"v0.22.0","url":"https://github.com/denoland/deno/releases/tag/v0.22.0"},{"tag":"v0.21.0","url":"https://github.com/denoland/deno/releases/tag/v0.21.0"}]
```

I don't know if there is rate limiting on the GitHub release feed - I don't think so.

---
fixes #138

### DIFF
--- a/worker/api.js
+++ b/worker/api.js
@@ -1,0 +1,40 @@
+export async function handleAPIRequest(request) {
+  const url = new URL(request.url);
+  const accept = request.headers.get("accept");
+  const isJSON = accept && accept.indexOf("application/json") >= 0;
+
+  if (!isJSON) {
+    return new Response(
+      "The client does not accept 'application/json' content",
+      {
+        status: 400,
+        statusText: "Bad Request"
+      }
+    );
+  }
+
+  if (url.pathname == "/api/versions") {
+    const resp = await fetch("https://github.com/denoland/deno/releases.atom");
+    const atom = await resp.text();
+    const matches = [...atom.matchAll(/<title>v(.*)<\/title>/g)];
+    if (!matches) {
+      return new Response("Failed to fetch releases from GitHub", {
+        status: 500,
+        statusText: "Internal Server Error"
+      });
+    }
+    const versions = matches.map(match => ({
+      tag: "v" + match[1],
+      url: "https://github.com/denoland/deno/releases/tag/v" + match[1]
+    }));
+    return new Response(JSON.stringify(versions), {
+      status: 200,
+      headers: { "content-type": "application/json" }
+    });
+  }
+
+  return new Response("API method not found: " + url.pathname, {
+    status: 404,
+    statusText: "Not Found"
+  });
+}
--- a/worker/api.js
+++ b/worker/api.js
@@ -2,27 +2,63 @@ export async function handleAPIRequest(request) {
   const url = new URL(request.url);
 
   if (url.pathname == "/api/versions") {
-    const resp = await fetch("https://github.com/denoland/deno/releases.atom");
-    const atom = await resp.text();
-    const matches = [...atom.matchAll(/<title>v(.*)<\/title>/g)];
-    if (!matches) {
-      return new Response("Failed to fetch releases from GitHub", {
-        status: 500,
-        statusText: "Internal Server Error"
-      });
-    }
-    const versions = matches.map(match => ({
-      tag: "v" + match[1],
-      url: "https://github.com/denoland/deno/releases/tag/v" + match[1]
-    }));
-    return new Response(JSON.stringify(versions), {
-      status: 200,
-      headers: { "content-type": "application/json" }
-    });
+    return handleVersionRequest(request);
   }
 
   return new Response("API method not found: " + url.pathname, {
     status: 404,
     statusText: "Not Found"
+  });
+}
+
+const corsHeaders = {
+  "Access-Control-Allow-Origin": "*",
+  "Access-Control-Allow-Methods": "GET, HEAD, OPTIONS",
+  "Access-Control-Allow-Headers": "Content-Type"
+};
+
+async function handleVersionRequest(request) {
+  const { method } = request;
+  if (method === "OPTIONS") {
+    return new Response("200 OK", {
+      status: 200,
+      statusText: "OK",
+      headers: corsHeaders
+    });
+  }
+  // HEAD requests are handled by Cloudflare Workers
+  if (method !== "GET" && method !== "HEAD") {
+    return new Response("405 Method Not Allowed", {
+      status: 405,
+      statusText: "Method Not Allowed",
+      headers: corsHeaders
+    });
+  }
+
+  let matches = [];
+  try {
+    // Get versions from Github and cache them for 120 seconds if the request was a success.
+    const resp = await fetch("https://github.com/denoland/deno/releases.atom", {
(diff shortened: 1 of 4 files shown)


worker/api.js
```diff
@@ -0,0 +1,40 @@
+export async function handleAPIRequest(request) {
+  const url = new URL(request.url);
+  const accept = request.headers.get("accept");
+  const isJSON = accept && accept.indexOf("application/json") >= 0;
```
MarkTiedemann: I think it would be better to be more liberal here: `application/*` as well as `*/*` should be acceptable here.

Also, is it possible to use `str.includes(val)` rather than `str.indexOf(val) >= 0` in a worker?

worker/api.js
```diff
@@ -0,0 +1,40 @@
...
+
+  if (!isJSON) {
+    return new Response(
+      "The client does not accept 'application/json' content",
+      {
+        status: 400,
```
MarkTiedemann: I believe this should be a `406 Not Acceptable`.

worker/api.js
```diff
@@ -0,0 +1,40 @@
...
+      url: "https://github.com/denoland/deno/releases/tag/v" + match[1]
+    }));
+    return new Response(JSON.stringify(versions), {
+      status: 200,
+      headers: { "content-type": "application/json" }
+    });
+  }
+
+  return new Response("API method not found: " + url.pathname, {
+    status: 404,
```
MarkTiedemann: I think it's common to check the request path before doing content negotiation in HTTP servers. Here, a request with an improper content type and an unknown path would return 400 (406) rather than 404, but it should be the other way around.

worker/api.js
lucacasonato: So you're saying we shouldn't check the accept header at all - I think that makes sense as we only ever respond with JSON anyway.

> Also, is it possible to use str.includes(val) rather than str.indexOf(val) >= 0 in a worker?

Yes, I copied this code from [here](https://github.com/denoland/deno_website2/pull/180/files#diff-ebde6d112f6a6eee03c43a03067292b6L15), but I'll change it.

worker/api.js
```diff
@@ -0,0 +1,40 @@
...
+    return new Response(
+      "The client does not accept 'application/json' content",
+      {
+        status: 400,
+        statusText: "Bad Request"
+      }
+    );
+  }
+
+  if (url.pathname == "/api/versions") {
```
MarkTiedemann: Optionally, you could check whether the request method is `GET` (or `HEAD`, or `OPTIONS`) and, if not, return `405 Method Not Allowed`.

If the method is `HEAD`, you could return the headers only, without the body.

If the method is `OPTIONS`, you could list allowed request methods by responding with an `Allow: <methods>` header and handle CORS headers so this API can be called from a client in the browser.

worker/api.js
```diff
@@ -0,0 +1,40 @@
...
+    const resp = await fetch("https://github.com/denoland/deno/releases.atom");
```
MarkTiedemann: Fetching the releases from GitHub may fail (network error, 4xx or 5xx error) or timeout, in which case, you could respond with `502 Bad Gateway` or `504 Gateway Timeout`.

wrangler.toml
```diff
@@ -1,6 +1,6 @@
 # Cloudflare Worker configuration
-account_id = "895762025d37fc687ecd72d7cc80204a"
-zone_id = "c192cf8ac042c681023493c52edd44c8"
```
ry: Undo

worker/api.js
lucacasonato: This is resolved because I removed content-type checking.

lucacasonato:
> Is there any UI for this?

No - this is purely an API to make getting recent Deno versions easier.

lucacasonato:
> There's no tests and no UI - so this would be very fragile. I wouldn't know if it broke or why this code was there - so I can't land it like this.

We have no way of testing the CF worker at all right now, so I can't write a test for this that does a full integration test. I could add UI for this, but I don't know where and what the use for it would be. I guess maybe a 'latest version' display on the homepage? Not 100% on that though - don't want to add bloat.

This whole PR might also just be out of the scope of the website. I just opened to solve #138 - it might be better to just have this be something that a third party site provides.

ry:
Having the latest version on the home page might be nice... but it also seems like an unnecessary extra request. *shrug*

MarkTiedemann:
https://github.com/dollarshaveclub/cloudworker could be used for testing (but that seems like another PR to me).

worker/api.js
```diff
@@ -0,0 +1,64 @@
+export async function handleAPIRequest(request) {
+  const url = new URL(request.url);
+
+  if (url.pathname == "/api/versions") {
```
MarkTiedemann: nit `===`for consistency

worker/api.js
```diff
@@ -0,0 +1,64 @@
...
+  return new Response("API method not found: " + url.pathname, {
+    status: 404,
+    statusText: "Not Found"
+  });
+}
+
+const corsHeaders = {
+  "Access-Control-Allow-Origin": "*",
+  "Access-Control-Allow-Methods": "GET, HEAD, OPTIONS",
+  "Access-Control-Allow-Headers": "Content-Type"
```
MarkTiedemann: AFAIK, you don't need to whitelist `Content-Type`, it's a "CORS-safelisted request header" (https://developer.mozilla.org/en-US/docs/Glossary/CORS-safelisted_request_header), i.e. one of the headers that is always allowed.

MarkTiedemann:
This looks pretty good to me now. :) gj @lucacasonato!

lucacasonato:
I have added a latest version display to the website now:
image If the api fails to respond, a fallback message without the version is shown.

To test this locally, replace the REMOTE_URL in the worker with a netlify deploy preview URL (e.g. 
https://5e3744598c2f9b00077fa57f--denoland.netlify.com/) and run `wrangler preview`.

ry:
@lucacasonato Thanks for this patch - I understand the need for it - but it's just too brittle and niche. It's not something I want to take on at this point. Sorry for the slow review. Closing without merge.

PR #181, opened 2020-01-09
Case insensitive filter and copy code.

fakoua:
- Make filter in RegistryIndex case insensitive
- Allow Copy in code block


lucacasonato:
:+1:
Could you maybe put the code copy at the end of the text field and remove the extra space at the start of the codeblock? I think the copy button should appear when you hover over the textfield, floating in the top right.

src/component/CodeBlock.tsx
```diff
@@ -13,18 +17,79 @@ darkTheme["hljs-selection"] = {
...
-    <SyntaxHighlighter
-      style={darkMode ? darkTheme : lightTheme}
-      language={props.language || "js"}
-      showLineNumbers={props.showLineNumbers || false}
-      wrapLines={true}
-      lineProps={props.lineProps}
-    >
-      {props.value}
-    </SyntaxHighlighter>
+    <Container
```
lucacasonato: This container adds an extra margin on the left and right - I don't think we want that.

src/component/CodeBlock.tsx
```diff
@@ -13,18 +17,79 @@ darkTheme["hljs-selection"] = {
...
+      newState = Object.assign({}, state)
+      newState.copied = false;
+      setState(newState)
+    }, 300)
+  }
+
+  const [state, setState] = React.useState({
+    showCopy: false,
+    copied: false,
+  });
```
lucacasonato: Could you maybe replace this with two state objects to make the state handling slightly less verbose - you can then get rid of all of the `Object.assign`s.

```typescript
const [showCopy, setShowCopy] = React.useState(false);
const [copied, setCopied] = React.useState(false)
```

fakoua:
Feedback fixed!
Thanks

lucacasonato:
Looks great! Is it possible to add a timeout for fade-out? So the copy button will stick around for half a second before fading out? Also I think a slightly shorter fade in would work better.

fakoua:
Fixed, thanks

lucacasonato:
Great! Just one more nitpick - could you remove the fade in and just keep the fade out? Also hovering over the code block, then leaving and then entering again makes the copy button disappear.

This should fix both things:

```typescript
  function onContainerEnter() {
    clearTimeout(timerState.timerId);
    setShowCopy({showCopy: true});
  }

  function onContainerLeave() {
    clearTimeout(timerState.timerId);
    let timerId = setTimeout(() => {
      setShowCopy({showCopy: false});
    }, 500);
    setTimer({timerId: timerId})
  }
```

fakoua:
Many thanks, yes entering and leaving was buggy.

Issue fixed.

ry:
This seems like a lot of code for what it’s achieving...

fakoua:
@ry
You are right, could be simplified using css in some cases.

fakoua:
@ry ,
It shows all the fork sync since started, the affected files one 2:
1- RegistryIndex.js:
```ts
name.toLowerCase().includes(query.toLowerCase()) ||
      (DATABASE[name].desc || "").toLowerCase().includes(query.toLowerCase())
```
2- CodeBlock.tsx:
```ts
const useStyles = makeStyles(theme => ({
  copyButton: {
    position: 'relative',
    right: '5px',
    float: 'right'
  },
  hidden: {
    display: 'none'
  },
  container: {
    position: 'relative',
    paddingLeft: '0',
    paddingRight: '0'
  },
  checkIcon: {
    color: 'rgb(80, 161, 79)'
  }
}));

...

const [timerState, setTimer] = React.useState({
    timerId: 0,
  });

  const [showCopyState, setShowCopy] = React.useState({
    showCopy: false,
  });
  
  const [copiedState, setCopied] = React.useState({
    copied: false,
  });

  function onContainerEnter() {
    clearTimeout(timerState.timerId);
    setShowCopy({showCopy: true});
  }

  function onContainerLeave() {
    clearTimeout(timerState.timerId);
    let timerId = setTimeout(() => {
      setShowCopy({showCopy: false});
    }, 500);
    setTimer({timerId: timerId})
  }

  function onCopy() {
    setCopied({copied: true});
    setTimeout(() => {
      setCopied({copied: false});
    }, 1300)
  }

<Container
      onMouseEnter={onContainerEnter}
      onMouseLeave={onContainerLeave}
      className={classes.container}>
      <Zoom in={showCopyState.showCopy || copiedState.copied} style={{ transitionDelay: '50ms' }}>
        <CopyToClipboard
          text={props.value}
          onCopy={onCopy}>
          <IconButton .........

```

lucacasonato:
> This seems like a lot of code for what it’s achieving...

@ry I do think it is a really important feature though. Being able to copy like this is very, very handy.

---

@fakoua You can reduce the line count some more by replacing the state management to use direct primitive types and not objects.

```typescript
  const [timer, setTimer] = React.useState(0);
  const [showCopy, setShowCopy] = React.useState(false);
  const [copied, setCopied] = React.useState(false);

  function onContainerEnter() {
    clearTimeout(timer);
    setShowCopy(true);
  }

  function onContainerLeave() {
    clearTimeout(timer);
    let timerID = setTimeout(() => {
      setShowCopy(false);
    }, 500);
    setTimer(timerID)
  }

  function onCopy() {
    setCopied(true);
    setTimeout(() => {
      setCopied(false);
    }, 1300)
  }
```

fakoua:
> > This seems like a lot of code for what it’s achieving...
> 
> @ry I do think it is a really important feature though. Being able to copy like this is very, very handy.
> 
> @fakoua You can reduce the line count some more by replacing the state management to use direct primitive types and not objects.
> 
> 
>   const [timer, setTimer] = React.useState(0);
>   const [showCopy, setShowCopy] = React.useState(false);
>   const [copied, setCopied] = React.useState(false);
> 
>   function onContainerEnter() {
>     clearTimeout(timer);
>     setShowCopy(true);
>   }
> 
>   function onContainerLeave() {
>     clearTimeout(timer);
>     let timerID = setTimeout(() => {
>       setShowCopy(false);
>     }, 500);
>     setTimer(timerID)
>   }
> 
>   function onCopy() {
>     setCopied(true);
>     setTimeout(() => {
>       setCopied(false);
>     }, 1300)
>   }

- Code cleanup, 
Thanks

ry:
I’m away from my computer the next two days - I’ll review in detail when I’m back.

ry:
This seems to have two unrelated bits. I wonder if they could be split into two PRs?

lucacasonato:
@fakoua Maybe remove the case insensitive filtering from this PR and just keep the copy button in this PR. Then you can open an other PR for the case insensitive filtering.

fakoua:
Hi @ry , @lucacasonato 

This PR contains only the copy code feature, once merged I will PR the case insensitive feature, 
Many thanks.

PR #189, opened 2020-01-15
Prevent layout jumping on home page and reduced bundle size

lucacasonato:
I removed all of the unused languages for the syntax highlighter. This reduced the CodeBlock bundle from ~500MB to about ~200MB.

On the home page the layout jumped up to six times before (4x for when the badges load, 1x for the logo and 1x for the codeblocks).

I fixed this by giving the badges and logo explicit heights. I also statically added the CodeBlock bundle to the main page to stop that from jumping because of React suspense. The initial page load time is about the same and the total bundle size is actually smaller due to the reduced size of the CodeBlock bundle.

While working on the logo sizing I made the styling more responsive which makes the site look a lot less squashed on mobile in my opinion.

### DIFF
--- a/src/component/CodeBlock.tsx
+++ b/src/component/CodeBlock.tsx
@@ -1,9 +1,15 @@
 import React from "react";
-import SyntaxHighlighter, {
+import {
+  Light as SyntaxHighlighter,
   SyntaxHighlighterProps
 } from "react-syntax-highlighter";
 import lightTheme from "react-syntax-highlighter/dist/cjs/styles/hljs/atom-one-light";
 import darkTheme from "react-syntax-highlighter/dist/cjs/styles/hljs/atom-one-dark";
+import javascript from "react-syntax-highlighter/dist/esm/languages/hljs/javascript";
+import typescript from "react-syntax-highlighter/dist/esm/languages/hljs/typescript";
+import json from "react-syntax-highlighter/dist/esm/languages/hljs/json";
+import markdown from "react-syntax-highlighter/dist/esm/languages/hljs/markdown";
+import shell from "react-syntax-highlighter/dist/esm/languages/hljs/shell";
 import { useDarkMode } from "../hook/theme";
 
 lightTheme["hljs-selection"] = {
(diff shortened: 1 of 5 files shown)


src/component/CodeBlock.tsx
```diff
@@ -1,9 +1,15 @@
...
+  Light as SyntaxHighlighter,
   SyntaxHighlighterProps
 } from "react-syntax-highlighter";
 import lightTheme from "react-syntax-highlighter/dist/cjs/styles/hljs/atom-one-light";
 import darkTheme from "react-syntax-highlighter/dist/cjs/styles/hljs/atom-one-dark";
+import javascript from "react-syntax-highlighter/dist/cjs/languages/hljs/javascript";
+import typescript from "react-syntax-highlighter/dist/cjs/languages/hljs/typescript";
+import json from "react-syntax-highlighter/dist/cjs/languages/hljs/json";
+import markdown from "react-syntax-highlighter/dist/cjs/languages/hljs/markdown";
+import shell from "react-syntax-highlighter/dist/cjs/languages/hljs/shell";
```
ry: Why is this cjs?

src/component/CodeBlock.tsx
```diff
@@ -13,6 +19,12 @@ darkTheme["hljs-selection"] = {
   backgroundColor: "#3a404b" // https://github.com/atom/one-dark-ui/blob/master/styles/ui-variables.less#L32
 };
 
+SyntaxHighlighter.registerLanguage("javascript", javascript);
+SyntaxHighlighter.registerLanguage("typescript", typescript);
+SyntaxHighlighter.registerLanguage("json", json);
+SyntaxHighlighter.registerLanguage("markdown", markdown);
+SyntaxHighlighter.registerLanguage("shell", shell);
```
ry: We could safely get rid of shell. We don't have shell scripts - seems a waste to try to highlight one-line commands.

src/component/CodeBlock.tsx
lucacasonato: Otherwise tests don't work because jest + cra can't handle esm imports.

lucacasonato:
Whoops I wrote MB but meant KB. And I was talking about gzipped sizes from chrome network panel.

PR #193, opened 2020-01-18
Add SwissKnife

fakoua:
SwissKnife is a Deno module that allows you to do some useful tasks on Windows OS using NirCmd v2.86


fakoua:
order issue

PR #195, opened 2020-01-19
Change registry menu to horizontal

fakoua:
This is the PR1 to change the registry menu to Horizontal, 
Once merged I will PR the case sensitive issue.
Thanks

### DIFF
--- a/src/page/Registry.js
+++ b/src/page/Registry.js
@@ -1,5 +1,5 @@
 import React from "react";
-import { Box, Divider } from "@material-ui/core";
+import { Box, ButtonGroup, Button } from "@material-ui/core";
 import { useLocation } from "react-router-dom";
 import Link from "../component/Link.tsx";
 import Spinner from "../component/Spinner";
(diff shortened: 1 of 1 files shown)


src/page/Registry.js
```diff
@@ -107,25 +106,23 @@ export default function Registry() {
...
+            {hasDocsAvailable ? " | " : ""}<Link to={state.repoUrl}>Repository</Link>
+          </span>
         ) : null}
         {state.rawUrl ? (
-          <p>
-            <Link to={state.rawUrl}>Raw</Link>
-          </p>
+          <span>
+            &nbsp;|&nbsp;<Link to={state.rawUrl}>Raw</Link>
+          </span>
```
ry: We should try to use material-ui components rather than putting in custom text stuff like this

I'm not sure what exactly to use, but this looks appropriate: https://material-ui.com/components/dividers/#divider

See  `<Divider orientation="vertical" />`

fakoua: d01

You previously agreed on pipe sep., do you want to use Material UI component instead?

fakoua: Ok @ry 
I'm using <Divider orientation="vertical"/> now with some style, please review.

Thanks

ry:
@fakoua I agree to the look of it, but I want the website to avoid custom styles and complex UI as much as possible. I don't like all the CSS needed to make Divider work... Rather we should find some material UI that just works without customization. I've replaced your work with a Button. What do you think?

fakoua:
Hi,
Looks fine, but it gives the impression of a button not a link or menu. 
Frankly speaking, I liked the first one with pipe, however it's up to you, let me know if you want me to implement your design.

ry:
I prefer the pipe too, but I want to keep within the material UI design....

fakoua:
OK, tomorrow will try my best to check the material ui site.

fakoua:
Now im using ButtonGroup pure material UI without styles, 
Please review the new design, 
Change already PR-ed
d02

PR #205, opened 2020-01-22
feat: better page titles

### LINKED ISSUE #204: The title of top page doesn't look beautiful.
The top page's title is now `deno /`. This doesn't look ideal. This also affects the title in the google search result page.

スクリーンショット 2020-01-21 15 15 51

kt3k:
~~[react-helmet](https://github.com/nfl/react-helmet) is a library for setting the contents in `<head>` tag such as `<title>`, `<meta>`, etc.~~

~~This PR introduces react-helmet, and gives more sensible titles to each page.~~

This PR introduces `Title` component for setting the titles in the page components.

---
closes #204

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "react-apexcharts": "^1.3.3",
     "react-copy-to-clipboard": "^5.0.2",
     "react-dom": "^16.10.2",
+    "react-helmet": "^5.2.1",
     "react-markdown": "^4.2.2",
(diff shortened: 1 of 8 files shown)


kt3k:
@ry I dropped `react-helment` dependency, but I'd prefer to keep the title settings inside the page components to avoid the duplication of routing logic. So I created a simple component `Title` which does the similar job as what react-helmet does for the title.

PR #209, opened 2020-01-25
Add deno-servicr

astanciu:
Servicr is a graceful startup and shutdown manager for your deno services.

### DIFF
--- a/src/database.json
+++ b/src/database.json
@@ -221,13 +221,7 @@
     "owner": "maxuuell",
     "repo": "deno_init",
     "desc": "The executable to set up a deno project 🦕"
-  },
-   "servicr": {
-    "type": "github",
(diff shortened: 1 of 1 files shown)


astanciu:
I'm not sure why the tests are failing here..

zhmushan:
`expect(key.includes("-")).toBeFalsy();`
you should use `_` instead of `-`

src/database.json
```diff
@@ -221,6 +221,12 @@
     "owner": "maxuuell",
     "repo": "deno_init",
     "desc": "The executable to set up a deno project 🦕"
+  },
+   "deno_servicr": {
```
ry: How about just calling it servicr here? the deno is redundant.

Issue #210, opened 2020-01-26
Support line numbers


brandonkal:
Links such as this:
https://deno.land/x/lib/kite.ts:204:73
should function in the browser. This is handy as that is the format used in error messages that throw in the terminal.

(on a seperate note, it would be nice to see a Deno option to display these errors resolved to their actual on-disk location so clicking on them will open in the editor)

PR #214, opened 2020-01-28
Add support for colon-based line references

isAdrisal:
Addresses #210 

Currently, navigating to [https://deno.land/std/examples/echo_server.ts#L5](https://deno.land/std/examples/echo_server.ts#L5) in a browser will display syntax-highlighted code, with line 5 highlighted.

As pointed out in the issue above, the Deno CLI provides line number:column references in a different format like so: [https://deno.land/std/examples/echo_server.ts:5:8](https://deno.land/std/examples/echo_server.ts:5:8)

This PR uses the redirect Worker to modify the URL provided by the CLI to use the hash-style line references used by GitHub and the rest of the site's routing logic. Column references are discarded.

One issue -- I don't know how to actually test the worker locally. Any pointers?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,15008 @@
+{
+  "name": "deno_website2",
+  "version": "0.0.1",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@babel/code-frame": {
+      "version": "7.8.3",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.8.3.tgz",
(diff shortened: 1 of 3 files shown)


isAdrisal:
Hmm, testing with [https://cloudflareworkers.com/](https://cloudflareworkers.com/) this patch doesn't seem to work. The React router doesn't update the URL being used by the app, so the source code fetch fails. Not sure how to resolve 😕

ry:
I'm all for the feature if you can get it to work. I think these regex are pretty fiddly and need to be tested. I suggest you move the code to src/util and add add some unit tests. Then the tested function can be imported in the worker.

Unfortunately we don't currently have a way to run tests for the CF worker.

PR #215, opened 2020-01-29
WIP: Redirect with explicit git references in the registry

nayeemrmn:
Currently in the registry when a tag isn't given, it silently serves up master. This makes it explicitly redirect to the URL containing an `@<ref>` to encourage this important practice. Only for github database entries.

It first looks for the tag name attached to the [latest release](https://developer.github.com/v3/repos/releases/#get-the-latest-release). If the repo has no releases it falls back to getting the SHA for the latest commit.

`https://deno.land/std/http/server.ts` -> `https://deno.land/std@v0.31.0/http/server.ts`
`https://deno.land/x/oak/mod.ts` -> `https://deno.land/x/oak@918c1b/mod.ts`

WIP: Need to figure out rate limit stuff.

### DIFF
--- a/src/util/registry_utils.js
+++ b/src/util/registry_utils.js
@@ -1,7 +1,7 @@
 import assert from "assert";
 import DATABASE from "../database.json";
 
-export function proxy(pathname) {
+export async function proxy(pathname) {
   if (pathname.startsWith("/std")) {
     return proxy("/x" + pathname);
   }
@@ -12,7 +12,7 @@ export function proxy(pathname) {
   const [nameBranch, ...rest] = nameBranchRest.split("/");
   const [name, branch] = nameBranch.split("@", 2);
   const path = rest.join("/");
-  const entry = getEntry(name, branch);
+  const entry = await getEntry(name, branch);
   if (!entry || !entry.url) {
     return null;
   }
@@ -27,12 +27,13 @@ export function proxy(pathname) {
  * @param  {string}                branch
  * @return {import('./types').Entry}
  */
-export function getEntry(name, branch = "master") {
+export async function getEntry(name, branch = null) {
   // denoland/deno_std was merged into denoland/deno. For a while we will try
   // to maintain old links for backwards compatibility with denoland/deno_std
   // but eventually tags before v0.20.0 will break.
   if (
     name === "std" &&
+    branch != null &&
     (branch === "v0.7.0" ||
       branch === "v0.8.0" ||
       branch === "v0.9.0" ||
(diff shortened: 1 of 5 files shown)


nayeemrmn:
I don't know that this is implemented properly with respect to async stuff.

ry:
@nayeemrmn This is a cool feature - pretty similar to how docs.rs works: https://docs.rs/rusty_v8

But I also think it should be valid to link to master branch. During prototyping this is usually what you want - this patch makes it difficult to get a link to master branch.

What if instead we displayed a warning on the master branch pages and give a link to the latest version - rather than redirecting?

nayeemrmn:
I think it's important for things like the manual to point to latest release. e.g. https://github.com/denoland/deno/issues/3987. Maybe this is less of a problem when there is stability but in general the latest release is the "correct" version.

nayeemrmn:
Also redirecting to some canonical URL for any given module is essential to avoiding cache duplication, even if it was master by default.

bartlomieju:
> @nayeemrmn This is a cool feature - pretty similar to how docs.rs works: https://docs.rs/rusty_v8
> 
> But I also think it should be valid to link to master branch. During prototyping this is usually what you want - this patch makes it difficult to get a link to master branch.
> 
> What if instead we displayed a warning on the master branch pages and give a link to the latest version - rather than redirecting?

Can `std@master` could work? I think that defaulting to master is pretty bad idea, it's especially obvious after deleting `std/testing/mod.ts`

src/util/registry_utils.js
```diff
@@ -91,3 +99,29 @@ export function getEntry(name, branch = "master") {
   }
   return null;
 }
+
+async function getGithubLatestVersion(owner, repo) {
+  const response = await fetch(
+    `https://api.github.com/repos/${owner}/${repo}/releases/latest`
+  );
```
lucacasonato: We can't do this from the worker as we would very quickly run into Github API rate limits. Limits are 60 req per hour for non authenticated users, and 3000 for authenticated users. Because we share the ip range where the worker is requesting from with other people (because that's how CF workers work), we will be through this limit permanently. IDK what the current statistics for deno.land/x are, but definitely more than 60 req per hour.

The only reason we can call the API for the website without running into limits is because we do this from the client.

I don't see any other way of doing this though, so it might be worth investigating if Github is willing to raise our rate limit to enable this - idk if there is precedent for that though.

hayd: isn't the hack to parse the releases rss feed: e.g. https://github.com/denoland/deno/releases.atom

lucacasonato: Yup. We might want to consult with Github first to see what they think is the right way to go here - I presume this API request is cheaper for them than requesting this large XML file. It might be in their own best interest to give us an increased rate limit here.

nayeemrmn: > isn't the hack to parse the releases rss feed: e.g. https://github.com/denoland/deno/releases.atom

I can't fetch that page from another site due to the CORS policy, try `await (await fetch("https://github.com/denoland/deno/releases.atom")).text()`. Know a good workaround?

lucacasonato: Nope - can't really easily work around CORS. You can proxy via the worker maybe?

lucacasonato:
So whats the plan for this PR?

We either need to proxy the `.atom` file via the worker so the client can access it without CORS (but I don't know if there is a hidden rate limit on that endpoint), or we use the GitHub API. In this case we should also add some caching using Cloudflare Workers KV so the amount of requests we send to GitHub stays consistently low even if our req counts go up. We should be able to limit the GitHub API calls to a maximum of _amount of registry entries_ * 60 so currently about 8520 maximum calls per hour if the cache duration is 1 minute. If we increase the cache time by 5x it decreases the amount of calls by 5x to 1704 (which would fall into a regular API key limit of 5000/req/hour - even with discrepancies). This would allow us to grow the registry to somewhere around 400 GItHub modules before we run into rate limiting.

What do you think?

/cc @ry

ry:
I welcome a patch to display a warning on pages that pull from the master branch.

PR #236, opened 2020-02-10
Add 3rd party library fp-ts

pe3:
Related issue under gcanti/fp-ts https://github.com/gcanti/fp-ts/issues/1130

### DIFF
--- a/src/database.json
+++ b/src/database.json
@@ -367,12 +367,12 @@
     "path": "src/",
     "desc": ""
   },
-   "fp-ts": {
+   "fp_ts": {
     "type": "github",
     "owner": "gcanti",
     "repo": "fp-ts",
     "path": "src/",
-    "desc": "Functional programming in TypeScript "
+    "desc": "Functional programming in TypeScript"
   },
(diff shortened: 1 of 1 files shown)


pe3:
Running `yarn test` locally it seems that database names with dashes are not allowed: 

      23 | test("database names with dashes are not allowed", () => {
      24 |   for (const key in DATABASE) {
    > 25 |     expect(key.includes("-")).toBeFalsy();
         |                               ^
      26 |   }
      27 | });
      28 | 

Is there any way for adding fp-ts to the database?

pe3:
I renamed the database entry: fp-ts -> fp_ts

pe3:
I wasn't aware of any method to test it the database entry actually works for Deno CLI so I don't know if it works.

tmueller:
Maybe this should be removed again. Since fp-ts is not a deno module, it won't load.

pe3:
@tmueller it should be removed. I already made a PR for removing it https://github.com/denoland/deno_website2/pull/239

PR #243, opened 2020-02-19
Update database.json

ebebbington:


ebebbington:
Making this pull request as the original maintainer of the repo/module seems to have stopped maintaining it, so i plan on fixing the main issue with it due to wanting to use the module myself

PR #244, opened 2020-02-19
Add deno-watch-v2 and deno-task-runner-v2

ebebbington:
The original versions use outdated Deno syntax and thus are broken. The owner seems very unlikely to fix them so I have tried fixing them myself

### DIFF
--- a/src/database.json
+++ b/src/database.json
@@ -785,6 +785,12 @@
     "repo": "deno-task-runner",
     "desc": "deno-task-runner"
   },
+  "task_runner": {
+    "type": "github",
+    "owner": "ebebbington",
+    "repo": "deno-task-runner-v2",
+    "desc": "deno-task-runner-v2"
+  },
   "tiny_templates": {
     "type": "github",
     "owner": "zekth",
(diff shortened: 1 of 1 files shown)


ebebbington:
Updated both projects to succeed on testing, although deno-task-runner fails because it's still pointing to the broken module on deno.land. I need to add a working deno-watch-v2 to get everything going again

ebebbington:
I'm fairly new to this - are the CI checks failing due to my projects?

src/database.json
```diff
@@ -828,6 +834,12 @@
     "repo": "deno-watch",
     "desc": "A pure deno file watcher."
   },
+  "watch-v2": {
```
ry: CI is failing because deno.land/x doesn't allow dashes in the project name. Change this to watch_v2

If you click on the failing CI, you'll see this error message
```
      23 | test("database names with dashes are not allowed", () => {
      24 |   for (const key in DATABASE) {
    > 25 |     expect(key.includes("-")).toBeFalsy();
         |                               ^
      26 |   }
      27 | });
      28 | 
```

ebebbington: Alright cheers, ill sort it now - I mistakenly thought that error was something unrelated!

PR #246, opened 2020-02-19
Fixed README preview for GitHub module subdirectories

### LINKED ISSUE #245: Wrong readme is displayed for subdirectories
The wrong readme is displayed under the directory view of subdirectories (e.g. https://deno.land/std/examples/)

lucacasonato:
Fixes #245

### DIFF
--- a/src/page/Registry.js
+++ b/src/page/Registry.js
@@ -217,7 +217,7 @@ async function renderDir(pathname, entry) {
       const rawUrl =
         entryType === "esm"
           ? `${entry.url}${path}README.md`
-          : `${entry.url}README.md`;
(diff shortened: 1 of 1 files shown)


src/page/Registry.js
```diff
@@ -217,7 +217,7 @@ async function renderDir(pathname, entry) {
       const rawUrl =
         entryType === "esm"
           ? `${entry.url}${path}README.md`
-          : `${entry.url}README.md`;
+          : `https://raw.githubusercontent.com/${owner}/${repo}/${entry.branch}/${path}README.md`;
```
ry: What about non-github entries?

lucacasonato: This is already in a `if (entryType === "github" || entryType === "esm")` block (line 176) so it must be `github` because its not `esm`.

PR #248, opened 2020-02-19
Fixed all trailing slash errors

### LINKED ISSUE #116: Module URL with no trailing slash has broken links
Example:

- https://deno.land/x/dirname/ (with slash)
    - mod.ts links to https://deno.land/x/dirname/mod.ts (good)
- https://deno.land/x/dirname (without slash)
    - mod.ts links to https://deno.land/x/mod.ts (bad)

Additionally, the effect of a bad link is a blank screen with no 404 error or any other message.

Example:

- https://deno.land/x/nonexistent/path/name

Expected result: 404 error

What is visible: A quick flash of a header and a blank screen.

### LINKED ISSUE #230: [bug] if directory path does not end in "/" links are broken
Example. Click on README.md for both locations
Works:
https://deno.land/x/lib/
Broken
https://deno.land/x/lib

lucacasonato:
Fix #230 and fix #123 and fix #116

Examples:
- /std -> /std/
- /x/abc -> /x/abc/
- /std/bytes -> /std/bytes/

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -42,14 +42,26 @@ function App() {
             <Route path="/style_guide(.html)?">
               <Redirect to="/std/style_guide.md" />
             </Route>
+            <Route path="/std/" component={Registry} strict />
             <Route path="/std/:stdPath" component={Registry} />
-            <Route path="/std/" component={Registry} />
+            <Route path="/std@:stdVersion/" component={Registry} strict />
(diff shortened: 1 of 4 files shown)


lucacasonato:
> It would be cool to have a test for these...

Added some tests. I am not sure how reliable the tests will be because they rely on the Github API. And that might be rate-limited in the runner.

PR #253, opened 2020-02-29
Folder restructuring and typescript+testing for worker

lucacasonato:
Moved all React stuff to //web and all worker things to //worker. Each have their own testsuite that gets run by GitHub Actions.


lucacasonato:
@ry The 'Request review from <user>' button is not appearing for me on this repo. I don't quite know why. Is that a setting you have to enable?

ry:
Review coming shortly - sorry for the delay. I need to try this out locally.

lucacasonato:
> Review coming shortly - sorry for the delay. I need to try this out locally.

No problem - take your time. Just seems strange that that button is not appearing... might have to email GitHub support.

worker/package.json
```diff
@@ -1,8 +1,31 @@
...
-  "license": "MIT"
+  "version": "1.0.0",
+  "description": "Cloudflare worker for deno.land",
+  "license": "MIT",
+  "scripts": {
+    "build": "webpack",
+    "dev": "NODE_ENV=development npm run build",
+    "format": "prettier --write 'src/**/*.{ts,js,css,json,md}'",
+    "test:clean": "rimraf ./transpiled/shared ./transpiled/worker",
+    "test": "npm run test:clean && npm run transpile && mocha --require source-map-support/register --recursive transpiled/worker/test",
```
ry: You use yarn.lock but call `npm run` here. I feel like we should use one or the other but not both.

ry:
Getting this warning for the web:
```
Compiled with warnings.

./node_modules/typescript/lib/typescript.js
Critical dependency: the request of a dependency is an expression
```

.editorconfig
```diff
@@ -1,20 +0,0 @@
...
-
-insert_final_newline = true
-
-[*.yml]
-indent_style = space
-indent_size = 2
-trim_trailing_whitespace = true
-
-[*.md]
-trim_trailing_whitespace = false
```
ry: I don't use this - so I don't particularly care about it - but why remove?

.eslintrc.json
```diff
@@ -1,41 +0,0 @@
...
-    "no-unused-vars": ["off"],
-    "react/no-unescaped-entities": ["error", { "forbid": [">", "}"] }],
-    "react/prop-types": ["off"]
-  },
-  "env": {
-    "browser": true,
-    "node": true,
-    "es6": true
-  }
-}
```
ry: Why remove the eslint?

.github/workflows/ci.yml
```diff
@@ -3,28 +3,51 @@ name: ci
...
         with:
           node-version: 12
       - run: yarn --frozen-lockfile
+        working-directory: ${{env.working-directory}}
       - run: yarn run lint
+        working-directory: ${{env.working-directory}}
       - run: yarn format --check
+        working-directory: ${{env.working-directory}}
       - run: CI=true yarn test
+        working-directory: ${{env.working-directory}}
```
ry: It would be less code if you just wrote
```
- run: CI=true yarn test
        working-directory: web
```

web/src/util/registry_utils.ts
```diff
@@ -1,7 +1,72 @@
...
+}
+
+interface URLDatabaseEntry {
+  type: "url";
+  name: string;
+  branch: string;
+  url: string;
+  repo: string;
+  raw: RawURLDatabaseEntry;
+}
```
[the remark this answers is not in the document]
lucacasonato: These are indeed new. I moved registry_utils to typescript. I'll split this out into a different PR.

.editorconfig
lucacasonato: I can re-add it, but usually editor configs should be git ignored I feel - they are not project related, only user preference right?

ry: 🤷🏻‍♂️
If it helps people contribute I think it’s fine. Like I said, I don’t use this, so I don’t particularly care :)

lucacasonato:
I'll split this into multiple PRs as requested.

PR #267, opened 2020-03-12
Update server example on homepage

ry:


src/page/Home.js
```diff
@@ -4,7 +4,7 @@ import CodeBlock from "../component/CodeBlock";
 import Title from "../component/Title";
 import "./Home.css";
 
-const code = `import { serve } from "https://deno.land/std@v0.30.0/http/server.ts";
+const code = `import { serve } from "https://deno.land/std@v0.35.0/http/server.ts";
```
nayeemrmn: v0.36.0?

ry: oops - fixed

hayd: Try with udd 🤪

PR #268, opened 2020-03-12
Shorten URLs to denofun modules

KSXGitHub:
Fix https://github.com/galkowskit/denofun/issues/53

/cc @galkowskit


galkowskit:
Yeah, that looks like a great idea from my side! :)

PR #256, opened 2020-03-06
Added my package.

AvexCodes:
Deno Chalk.

PR #269, opened 2020-03-12
Added DenoChalk

AvexCodes:
Added my module.

### DIFF
--- a/src/database.json
+++ b/src/database.json
@@ -228,6 +228,12 @@
     "repo": "deno",
     "desc": "A secure runtime for JavaScript and TypeScript."
   },
+   "deno_chalk": {
+    "type:" "github",
+    "owner": "avexcodes",
+    "repo": "denochalk",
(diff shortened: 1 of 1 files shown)


SkoshRG:
I suggest you check that the json you are trying to merge is valid, which in this commit isn't the case.


"deno_chalk": {
    "type": "github",
    "owner": "avexcodes",
    "repo": "denochalk",
    "desc": "Allows you to add colour, and styling to your console!"
},

This is what the object should be like. Also, I'm not sure what benefit this package adds? It seems quite exactly the same as the implementation in Deno's built in color formatting options, but with a less scalable API.

crowlKats:
duplicate PR? #256

kitsonk:
Also isn't it redundant with: https://github.com/denoland/deno/blob/master/std/fmt/colors.ts

ry:
stale

PR #272, opened 2020-03-13
Add KSXGitHub/deno-once as 'once'

KSXGitHub:
## Question

This repo contains only one file. Is it possible to specify `"path"` to that one file?


ry:
> This repo contains only one file. Is it possible to specify "path" to that one file?

Nope

PR #279, opened 2020-03-14
Refactor path structure

lucacasonato:
@ry Previously mentioned a webpack warning about  `Critical dependency: the request of a dependency is an expression`. This happens because Typescript calls require() with a variable rather than a static string. This only happens starting at version 3.7. I now just fixed the version to 3.6.


lucacasonato:
Closing because stale

Issue #274, opened 2020-03-14
Assertion Error on deno.land/x/denofun/


KSXGitHub:
**Step 1:** Go to https://deno.land/x/denofun/
**Step 2:** Open DevTools

ry:
Seems to still be an issue? Any idea what's causing this?

KSXGitHub:
This issue is just duplicate of https://github.com/denoland/deno_website2/issues/276 that I created by mistake.

KSXGitHub:
I have discovered the cause of the issue. It is because `"lib"` field in database.json not having trailing slash.

What fix would you prefer:
  * Editing `getEntry` function in `registry_utils.js` to automatically add trailing slash?
  * Or add a rule to `database.json` to force trailing slash?

bartlomieju:
> I have discovered the cause of the issue. It is because `"lib"` field in database.json not having trailing slash.
> 
> What fix would you prefer:
> 
> * Editing `getEntry` function in `registry_utils.js` to automatically add trailing slash?
> * Or add a rule to `database.json` to force trailing slash?

Second one is more explicit 👍

KSXGitHub:
@bartlomieju Choosing the second one means slowing down people from adding new package. I have experienced this when adding `deno-args`: I sorted the keys in wrong order, causing the CI to fail without informing me (via notification). I only discovered failure of CI few days later,

PR #280, opened 2020-03-15
Add rule that forces all paths to have trailing slash

KSXGitHub:
Fix https://github.com/denoland/deno_website2/issues/274


KSXGitHub:
@ry Maybe you should merge this too? This patch would help package developer avoid making the same mistake I made in the future.

PR #282, opened 2020-03-15
Add protobuf

seishun:
[Contributors welcome!](https://github.com/seishun/deno-protobuf)

I couldn't find any information about the js/experimental subdirectory in https://github.com/protocolbuffers/protobuf, but it looks like Google is intending for it to replace the existing runtime. It would be great if someone can get me in touch with the people working on it - I would love to find out more about the roadmap and if possible take part in the development.

### DIFF
--- a/src/database.json
+++ b/src/database.json
@@ -701,6 +701,13 @@
     "repo": "deno_process",
     "desc": "process module for Deno"
   },
+  "protobuf": {
+    "type": "github",
+    "owner": "seishun",
+    "repo": "deno-protobuf",
+    "path": "/runtime",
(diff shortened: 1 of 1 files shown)


src/database.json
```diff
@@ -701,6 +701,13 @@
     "repo": "deno_process",
     "desc": "process module for Deno"
   },
+  "protobuf": {
+    "type": "github",
+    "owner": "seishun",
+    "repo": "deno-protobuf",
+    "path": "/runtime",
```
seishun:
```suggestion
    "path": "runtime/",
```

Issue #296, opened 2020-04-02
issue(deno-debug): demo image not showing


rjoydip:
***Issue***
In reame file of `deno-debug` repo. There is a demo image but in deno docs it is not showing.

***Screenshot***
`deno-debug`:

![image](https://user-images.githubusercontent.com/15318294/78232765-5b3f3c80-74f2-11ea-8595-e7c096e5c014.png)

<hr />

`In deno dock`s:

![image](https://user-images.githubusercontent.com/15318294/78232706-45317c00-74f2-11ea-9ec4-46c561d5806d.png)

lucacasonato:
This should be fixed.

Issue #289, opened 2020-03-22
docs broken for https://deno.land/x/drash/


zhmushan:
minimal:
```tsx
function Markdown(props: MarkdownProps) {
  if (!props.source) {
    return null;
  }
  return (
    <ReactMarkdown
      source={`<a></a><a></a>`}
      renderers={renderers}
      plugins={[toc]}
      escapeHtml={false}
      astPlugins={[parseHtml]}
    />
  );
}
```
If we remove `parseHtml`, the page works, and it looks no different

lucacasonato:
Can be closed because doc.deno.land

PR #297, opened 2020-04-02
htmlParser valid node refactor

rjoydip:
This PR fixed both #296 & #289 

No need to use allowTags array. It could be handled as per the docs [with-custom-processing-instructions](https://github.com/aknuds1/html-to-react#with-custom-processing-instructions).

### DIFF
--- a/src/component/Markdown.tsx
+++ b/src/component/Markdown.tsx
@@ -22,42 +22,10 @@ interface HeadingRendererProps {
   level?: any;
 }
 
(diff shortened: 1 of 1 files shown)


rjoydip:
@lucacasonato Thanks for the suggestion 😃 . I will try to fix this in different way.

ry:
As @lucacasonato said. Thanks for the contribution anyway. Closing without merge.

PR #311, opened 2020-04-07
Fix typedoc redirect

lucacasonato:
I messed up in my PR yesterday which broke the /typedoc links. Our function `redirect` in the worker actually proxies and does not redirect which confused me. /typedoc should now correctly redirect.

### DIFF
--- a/worker/index.js
+++ b/worker/index.js
@@ -18,12 +18,10 @@ async function handleRequest(request) {
   const maybeProxyElsewhere =
     url.pathname.startsWith("/std") || url.pathname.startsWith("/x");
 
-  // TODO(ry) Support docs without hitting S3...
   if (url.pathname.startsWith("/typedoc")) {
-    return redirect(
-      url,
+    return Response.redirect(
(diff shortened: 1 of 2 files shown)


wrangler.toml
```diff
@@ -1,5 +1,5 @@
 # Cloudflare Worker configuration
-account_id = "895762025d37fc687ecd72d7cc80204a"
+account_id = "93bfcf72dfea0e9f2bf5563bc71052cf"
```
ry, line 2: Undo

worker/index.js
```diff
@@ -18,12 +18,10 @@ async function handleRequest(request) {
...
     url.pathname.startsWith("/std") || url.pathname.startsWith("/x");
 
-  // TODO(ry) Support docs without hitting S3...
   if (url.pathname.startsWith("/typedoc")) {
-    return redirect(
-      url,
+    return Response.redirect(
       "https://doc.deno.land/https/github.com/denoland/deno/releases/latest/download/lib.deno.d.ts",
-      request
+      307
```
ry, line 24: Maybe instead of redirecting we should continue to serve up the old typedoc.

PR #315, opened 2020-04-09
Auto-focus Third Party Modules search field

chances:
Add the [`autoFocus`](https://material-ui.com/api/text-field/#props) attribute to the Third Party Modules search field for better ergonomics when navigating directly to [deno.land/x](https://deno.land/x).

Related to #176.

### DIFF
--- a/src/page/RegistryIndex.js
+++ b/src/page/RegistryIndex.js
@@ -66,6 +66,7 @@ export default function RegistryIndex() {
         label="Search"
         type="search"
         value={query}
+        autoFocus={true}
         onChange={e => setQuery(e.target.value)}
       />
 
--- a/src/page/RegistryIndex.js
+++ b/src/page/RegistryIndex.js
@@ -66,8 +66,8 @@ export default function RegistryIndex() {
         label="Search"
         type="search"
         value={query}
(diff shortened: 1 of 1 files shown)


lucacasonato:
This is bad for mobile: https://github.com/denoland/deno_website2/pull/177#issuecomment-572389693, https://github.com/denoland/deno_website2/pull/198

chances:
I can only reproduce the "bad" behavior described in [#177 (comment)](https://github.com/denoland/deno_website2/pull/177#issuecomment-572389693) in the iPhone Simulator (iPhone 8, iOS 13). The keyboard is _not_ shown in Chrome on my Pixel 3a device, nor in the Simulator (iPhone 11, iOS 13).

To mitigate the issue can we consider moving the search field farther above the fold? Like immediately below the "Third Party Modules" header? What about only enabling `autoFocus` on desktops?

lucacasonato:
It does a really obnoxious zoom + scroll on iOS. Desktop only should be fine though.

ry:
Closing because stale and @lucacasonato had issues with it. We appreciate the effort @chances but closing without merge.

PR #331, opened 2020-04-16
(Hopefully) fixed README.md casing bug

Schotsl:
I don't have any experience developing in React, although I do have experience with Vue.js and a lot of JavaScript. I wrote this "fix" to allow the site to display any readme.md file without caring about the capitalization!

I'm hoping someone with the repo already running on their machine could check if it actually works though since I don't have a live environment.

### DIFF
--- a/src/page/Registry.js
+++ b/src/page/Registry.js
@@ -228,7 +228,7 @@ async function renderDir(pathname, entry) {
       const rawUrl =
         entryType === "esm"
           ? `${entry.url}${path}${entry.name}`
-          : `https://raw.githubusercontent.com/${owner}/${repo}/${entry.branch}/${path}${entry.name}`;
+          : `https://raw.githubusercontent.com/${owner}/${repo}/${entry.branch}/${path}/${entry.name}`;
       try {
         const response = await fetch(rawUrl);
         if (response.ok) body = await response.text();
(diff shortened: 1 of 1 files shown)


Schotsl:
It appears something is going wrong while testing the code but I can't figure out why, maybe someone with more experience could help out :)

src/page/Registry.js
```diff
@@ -224,11 +224,11 @@ async function renderDir(pathname, entry) {
...
     // no useful files exist in the repository, so opt to attempt
     // showing the contents of the README file instead if it exists.
-    if (files.some(entry => entry.name === "README.md")) {
+    if (files.some(entry => entry.name.toLowerCase() === "readme.md")) {
       const rawUrl =
         entryType === "esm"
-          ? `${entry.url}${path}README.md`
-          : `https://raw.githubusercontent.com/${owner}/${repo}/${entry.branch}/${path}README.md`;
+          ? `${entry.url}${path}${entry.name}`
+          : `https://raw.githubusercontent.com/${owner}/${repo}/${entry.branch}/${path}/${entry.name}`;
```
lucacasonato, line 231: The reason the tests are failing is because the `entry` in `entry.name` here is not referring to the `entry` with with the `README.md`.

You have to store the file entry you find that has a name of README.md and use its name:

```js
    const readme = files.find(
      entry => entry.name.toLowerCase() === "readme.md"
    );
    if (readme) {
      const rawUrl =
        entryType === "esm"
          ? `${entry.url}${path}${readme.name}`
          : `https://raw.githubusercontent.com/${owner}/${repo}/${entry.branch}/${path}${readme.name}`;
```

Schotsl: Ahh, thank you very much! Now it passes the tests :)

PR #332, opened 2020-04-17
Add talk-lib

JipFr:
I've added talk-lib, which is a library I made to interact with Nextcloud Talk


src/database.json
```diff
@@ -1013,6 +1013,12 @@
     "repo": "take-five",
     "desc": "A minimal (no external dependency) REST framework"
   },
+  "talk-lib": {
```
lucacasonato, line 1016: Dashes are not allowed:
```suggestion
  "talk_lib": {
```

PR #337, opened 2020-04-20
i18next libraries

adrai:


src/database.json
```diff
@@ -708,12 +733,6 @@
     "repo": "nanoid",
     "desc": "A tiny, secure, URL-friendly, unique string ID generator for JavaScript."
   },
-  "nessie": {
-    "type": "github",
-    "owner": "halvardssm",
-    "repo": "deno-nessie",
-    "desc": "A database migration tool for deno."
-  },
```
ry, line 716: Why is this being removed?

adrai: ohhh... this was not my intension... will fix imediately

adrai: should, be back...
sorry, I really don't know why this happened

PR #343, opened 2020-04-23
Centered spinner in viewheight

iDavidB:
It bothered me a lot that the spinner was always on the top left, especially on the homepage where it doesn't show any breadcrumbs while loading.


src/component/Spinner.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+export default function Spinner() {
+  const classes = useStyles();
+
+  return (
+    <div className={classes.spinner}>
+      <CircularProgress />
+    </div>
+  );
+}
```
lucacasonato, line 21: Can't you just center using flexbox? `style={{display: "flex", justifyContent: "center"}}` on the wrapper `div`. This adds a lot of JS processing which really isn't needed.

iDavidB: I used flex box initially but noticed the spinner still being off-center when something else is rendered (The breadcrumb). This being absolutely positioned will make sure it's always perfectly centered. I could apply the styling in-line if that is preferred though.

ry:
Thanks for the patch! I'll wait for @lucacasonato's approval.

src/component/Spinner.tsx
lucacasonato: Ah, makes sense then. :+1:

PR #379, opened 2020-05-10
added denoenv

rubiin:
Dotenv module port of nodejs dotenv

### DIFF
--- a/src/database.json
+++ b/src/database.json
@@ -325,6 +325,12 @@
     "repo": "deno_search",
     "desc": "A CLI tool to search for third party packages on deno.land"
   },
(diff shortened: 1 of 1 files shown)


crowlKats:
_another one_

rubiin:
> _another one_

I was using yours module earlier but i didnt find an option

crowlKats:
ah i see. you could have just opened an issue, i would have addressed it ASAP

rubiin:
> ah i see. you could have just opened an issue, i would have addressed it ASAP

I will open an issue about it. Seems like it would be very helpful option for your module in the long run

PR #382, opened 2020-05-10
added memoize lib

rubiin:
A simple and easy to use memoize lib for deno

### DIFF
--- a/src/database.json
+++ b/src/database.json
@@ -779,13 +779,13 @@
     "type": "github",
     "owner": "oakserver",
     "repo": "media_types",
-    "desc": "A simple and easy to use memoize lib for deno."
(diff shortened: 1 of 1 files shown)


src/database.json
```diff
@@ -779,6 +779,12 @@
     "type": "github",
     "owner": "oakserver",
     "repo": "media_types",
+    "desc": "A simple and easy to use memoize lib for deno."
```
JamesBroadberry, line 782: It looks like the description has been added to the wrong place (swapped with the media types module)

rubiin: sure i will give it a check,

rubiin: done 

> It looks like the description has been added to the wrong place (swapped with the media types module)

done

PR #386, opened 2020-05-11
feat: add selector for versions

MariusVatasoiu:
This PR adds a dropdown for versions on /std and /x pages.
- It displays master and all tags of the repo
- Selected value based on version from URL
- On change redirects to selected version
- For /std it displays only tags that starts with std/


ry:
As discussed in discord, we've actually got a new website coming out very soon now which includes this feature. I appreciate the contribution @MariusVatasoiu - I apologize for the unmerged effort.

PR #396, opened 2020-05-12
update to mention scoop support

erikyuzwa:
`scoop install deno`
was supported via this PR - https://github.com/lukesampson/scoop/pull/2570/files


erikyuzwa:
if I run `yarn format` then about 20 files are updated. I can commit that to this PR, just checking if that's normal or a "my machine" thing.

lucacasonato:
Thanks for the contribution, but this was fixed by #408.

erikyuzwa:
noice - thanks @lucacasonato

PR #418, opened 2020-05-14
Move relative imports to absolute imports

hswolff:
Supported now in next 9.4 https://nextjs.org/blog/next-9-4#absolute-imports-and-aliases

Happy to iterate on this any way you please.

Or...even if you despise these changes also happy to reject them. 😄


tsconfig.json
```diff
@@ -12,8 +12,9 @@
     "moduleResolution": "node",
     "resolveJsonModule": true,
     "isolatedModules": true,
-    "jsx": "preserve"
+    "jsx": "preserve",
+    "baseUrl": "."
```
hswolff, line 16: I tried to set this to `"baseUrl": ".."` in `worker/tsconfig.json` but it did not work.

Perhaps you can't have a baseUrl in a parent directory?

hswolff:
Going to re-investigate this via yarn workspaces...

PR #421, opened 2020-05-14
Add deno_logo.png & move deno_logo_4.gif to public/images

searchableguy:
This change fixes links in https://github.com/denoland/deno/blob/master/docs/introduction.md
The old logo was removed somehow so I got it from the web archive and tried to minimize the changes to avoid breaking old links.


lucacasonato:
Thanks :-D

PR #440, opened 2020-05-14
Add is-promise npm package

ForbesLindesay:
It can currently be used in deno via:

```js
import isPromise from 'https://unpkg.com/is-promise@4.0.0/index.mjs';

if (isPromise({then: () => {}})) {
  console.log('it is a promise');
} else {
  console.log('it is not a promise');
}
```

I'm not sure if there's anything else I should be doing to properly support deno.

### DIFF
--- a/database.json
+++ b/database.json
@@ -720,6 +720,11 @@
     "repo": "is_exe",
     "desc": "detect if a file is executable or not"
   },
+  "is-promise": {
+    "type": "npm",
+    "package": "is-promise",
+    "desc": "Test whether an object looks like a promise"
+  },
   "jbq": {
     "type": "github",
     "owner": "krnik",


ForbesLindesay:
@ry is that an intentional choice not to allow `-` in the module name? It seems like that's quite a significant restriction for people moving from node.

lucacasonato:
Yes that is an intentional choice

ry:
Closing because red. Start a new PR if you want to get this in.

ForbesLindesay:
Given that you do not support `-` in package names, I don't really see a path forwards here. In addition, the number of people who simply down-voted the PR, leads me to conclude that deno is a toxic community. I will not be supporting deno in my packages for the foreseeable future. I do sincerely hope both problems can be fixed, as there's a lot to like about deno.

nayeemrmn:
@ForbesLindesay I didn't down-vote but to provide context, the recent history of your package renewed a sentiment amongst the community that Deno would be a chance to address some problems with the existing ecosystem, particularly discouraging trivial dependencies.

Maybe it will, maybe it won't. But down-voting packages like this is unfortunately the only way people have of steering towards the first case.

I agree about `-` given that we have an `"npm"` type. I disagree that the community is toxic for simply down-voting 🤷‍♂️

Pointotech:
In which world is `-` not a valid URL character? It's literally the only idiomatic way to separate words in a URL. This is a really weird decision to make, especially without justifying it, and especially in an ecosystem like Deno where ALL PACKAGES are loaded from URLs.

Why is the Deno community attempting to get rid of the harmless dash character, and slow down interop/migration from Node for no apparent reason?

Soremwar:
@Pointotech By definition `-` isn't an alphanumeric character. `_` is.

lucacasonato:
@Pointotech We allow only a single separator character (`-`). This is to avoid many modules having a similar name: e.g. `is_even` and `is-even`. The decision was made over a year ago, and won't me changed. If you don't like deno.land/x because of this, host your packages somewhere else. The beauty of standard URL imports is that you are not forced to use any specific registry like deno.land/x :-)

andreespirela:
@ForbesLindesay not that it's relevant, but I hardly disagree with your statement:  
> leads me to conclude that deno is a toxic community
You should really give it a try to our Discord chat.

Pointotech:
That last comment is hardly fair, though. Every Discord chat is a toxic mess, because of Discord.

PR #444, opened 2020-05-14
Add package: create-denyo

MrPeker:
Easily install deno with a single command and run a sample server 🔮


bartlomieju:
Isn't it a Node package that allows to install and run Deno? I don't think you could use it with Deno itself

PR #449, opened 2020-05-14
Better manual docs title

MrPeker:
Add manual docs title to browser title for better navigation between open manual tabs. 

### Why?

When you open a few new tabs on your browser they will appear like this:  
 

This makes harder navigate between manuals when you learning. So, I added manual docs title to browser title for better navigation.

### DIFF
--- a/components/Manual.tsx
+++ b/components/Manual.tsx
@@ -102,7 +102,7 @@ function Manual() {
   useEffect(() => {
     // Is doc Markdown loaded?
     if (content && content.length > 0) {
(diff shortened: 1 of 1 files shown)


bartlomieju:
@lucacasonato please review

lucacasonato:
I agree that this should be done, but I don’t think it’s the right way. You should use nextjs’s Head component and use the title from the table of contents rather than the top heading of the markdown.

ry:
I'm going to close this PR because it's old. Thanks for the patch and getting the conversation started @MrPeker.

PR #450, opened 2020-05-14
Added readline

vslinko:


lucacasonato:
@vslinko FYI: https://deno.land/std/io/mod.ts contains this functionality already

vslinko:
@lucacasonato ow, I looked for this function in std/io but somehow I've missed it.

Looks like `readSlice` have significant difference from my module — it expects single byte delimiter. Also `readline` provides simple and pretty API.

vslinko:
I'm going to develop csv reader module based on `readline`.

vslinko:
I think that search for function definitions from `std` and external modules could be convenient for deno.land website.

Issue #453, opened 2020-05-14
Consider adding background-color on `search_field`


sidvishnoi:
This is how the search field looks on Firefox on Ubuntu.
![image](https://user-images.githubusercontent.com/8426945/81967278-549afd80-9638-11ea-822c-c9c63341f19e.png)
vs in Chrome:
![image](https://user-images.githubusercontent.com/8426945/81967358-6d0b1800-9638-11ea-9aa2-db763eff6168.png)

~I would've sent a PR, but I don't want to learn Tailwind at this time.~

Trivia: Firefox uses a special color value called `-moz-Field` as input `background-color`, which is based on OS themes, but Chrome has `#ffffff` by default.

**Edit:** Oh wait. It's just `.bg-white` :sweat_smile:

kennanseno:
@bartlomieju can be closed

PR #454, opened 2020-05-14
Add background color to search input

sidvishnoi:
Fixes https://github.com/denoland/deno_website2/issues/453


ry:
Thanks but we recently removed the search field

PR #456, opened 2020-05-14

kennanseno:
Not showing anything on dir paths for now rather than 0B which is misleading.

Realistically we should pull total size inside a directory but we've bigger fish to fry.


ry:
Closing because red. Please start a new PR if you want to get this in still.

PR #460, opened 2020-05-14
Add denodate to database.json

t-heu:
no description for now, is still in development


bartlomieju:
@t-heu thanks, but somethings wrong with diff - it shouldn't have a few thousand lines, please make sure that formatting is not changed

PR #468, opened 2020-05-14
Add path resolver

halvardssm:
This module contains a resolver for getting a path from an array of paths.


halvardssm:
Realized that it was one hidden in std, never mind :P

PR #486, opened 2020-05-15
Add typescript to database

parzhitsky:
Add `typescript` to database.json

### DIFF
--- a/database.json
+++ b/database.json
@@ -1450,6 +1450,12 @@
     "repo": "deno_types",
     "desc": "Definition files for JavaScript libraries compatible with Deno"
   },
+  "typescript": {
+    "type": "github",
+    "owner": "microsoft",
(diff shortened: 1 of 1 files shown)


database.json
```diff
@@ -992,12 +986,6 @@
     "repo": "deno-nessie",
     "desc": "A database migration tool for deno."
   },
-  "nessie": {
-    "type": "github",
-    "owner": "halvardssm",
-    "repo": "deno-nessie",
-    "desc": "A database migration tool for deno."
-  },
```
ry, line 1000: don't remove these

parzhitsky: They are duplicate keys though; is this intended?

image
image

ry: not related to this change

parzhitsky: Duplicates removed in #497

Issue #470, opened 2020-05-14
Search is not working in the Manual


timharek:
I'm using latests Firefox 76.0.1, also tried latest Safari and no luck with the search function in the manual; https://deno.land/manual

qadratic:
I doubt there is some code to do that. Its very much work in progress.

lucacasonato:
Yup - it is not implemented atm. We should remove the search field for the moment.

qadratic:
I can actually make it functional but at the moment I am opening a pr to disable it.

nayeemrmn:
@lucacasonato A single page view would be great, so we could <kbd>Ctrl</kbd>+<kbd>F</kbd> and for lots of other reasons. Might not need a built-in search. Would it be very difficult?

loilo:
I think even on a single page, a built-in search would still have benefits, especially when it comes to prioritization of search terms (e.g. higher weighted in headlines than in body text etc.)

lucacasonato:
> @lucacasonato A single page view would be great, so we could Ctrl+F and for lots of other reasons. Might not need a built-in search. Would it be very difficult?

Difficult no, just a lot of fetches to GitHub.

nayeemrmn:
We need a way to publish some set of auto-generated files with every commit on master. This would be used for `lib.deno.d.ts` (which can only be accessed on latest release right now, a crippling inconsistency), `docs/toc.json`, and a pre-concatenated single page manual.

We do this with https://github.com/denoland/benchmark_data...

lucacasonato:
Algolia DocSearch might work for this too: https://docsearch.algolia.com/

erlend-sh:
We’re big fans of Rust and Deno at [MeiliSearch](https://github.com/meilisearch/MeiliSearch) and we’d love to assist with setting up a self-hosted or managed search for you should the need arise.

https://docs.meilisearch.com/resources/howtos/search_bar_for_docs.html

lucacasonato:
Update: we have reached out to Algolia about DocSearch.

PR #492, opened 2020-05-15
removed non functional(atm) search bar

qadratic:
So I commented out the non functional search bar in Manual.tsx
meanwhile I am working on making a search functionality.

fixed issue: #470


ry:
cc @lucacasonato What were your plans for the search bar? I actually never noticed it until now.

qadratic:
okay so while typing, search results will be will be displayed on drop-down as suggestions. Each dropdown element will contain a link to that particular heading.

lucacasonato:
> What were your plans for the search bar? I actually never noticed it until now.

@ry Added it in the original design and wanted it to do full manual search, but never actually implemented it and then forgot to remove it again. Algolia DocSearch seems pretty great, maybe we can integrate that? https://docsearch.algolia.com/

lucacasonato:
@ry I guess, but DocSearch is really awesome :-D

Issue #406, opened 2020-05-13
Rethink the Third Party Modules registry


zjopy:
This issue has already been brought up multiple times in the old repo [#26](https://github.com/denoland/registry/issues/26),
[#117](https://github.com/denoland/registry/issues/117) and in this [issue comment](https://github.com/denoland/deno/issues/2473#issuecomment-623244622) on the main repo. In the meantime this new repo came up, and the old issues were archived. Now with v1 dropping and users starting to adopt Deno, this becomes more and more important, which is why I want to open an issue for it.

## The Problem

Modules can be added to the `database.json` to make them show up on the Third Party Modules page `https://deno.land/x/` and available under the URL `https://deno.land/x/MODULE_NAME@BRANCH/SCRIPT.ts`.

There are a few reasons why this is a suboptimal idea.

1. Scale

If Deno becomes even only half as popular as Node, there would be hundreds of modules added each day. It will totally flood the commit history and make the repo hard to maintain. Every PR needs to be verified by a maintainer, otherwise people could manipulate entries in any way. This is bad. Just looking at the recent commit history most of the commits are already related to adding entries for Third Party Modules.

2. Unique module names

Since module names must be unique, this will give rise to bad naming conflicts. Only the first module can be added, leaving anybody else with the same module name left out. This will be more like a domain name registry where all the good names are already taken. Even worse, entries for old unmaintained modules will block forever, because somebody out there might still depend on the URL so they can't be removed. This is even worse than domain names, which can at least expire.

3. GitHub only

The current system allows only links to GitHub registries. This might currently not be a problem, since GitHub is the defacto standard Git repository host, but inherently being locked into one host by choice of a too narrow system is never good. Also limiting the registry to GitHub is effectively centralizing it, which achieves the opposite effect of what Deno tries to achieve by using the ES module system.

In summary, the current module registry doesn't scale. It is high maintanance, requires unique module names, and is centralised.

## A solution

What people need is an overview of modules compatible with Deno and a convenient way to link to them.

Currently the Third Party Modules registry solves both, albeit in a very limited way as highlighted above. Entries in the `database.json` are used to build the overview, as well as generate the links available through `https://deno.land/x/...`.

I believe these two features shouldn't be made dependent on each other, but looked at individually.

- Module overview

There needs to be a way for developers to add (and remove) modules that scales well for lots of modules each day. Also it shouldn't be to the burden of the Deno maintainers, there should be an automated system of some sorts.

[...]

PR #501, opened 2020-05-15
Add types for database

parzhitsky:
Add type-safety and IntelliSense (editor suggestions) to database file


AnandChowdhary:
I don't know if this is a great idea, other projects may also rely upon the `database.json` file (like ours does) which this PR would break. I personally think it's much easier to consume JSON than an object in many use cases... :(

parzhitsky:
The fact that this is a breaking change is actually my biggest concern. Would you mind jotting about how exactly your project uses the json? Perhaps, it might be fixed by dynamically generating files at some point, or something like that (or something completely different 🙂).

As for ease of use, I tend to disagree, since a) with importing (or otherwise incorporating) the file you import both value (i.e., JavaScript implementation) _and_ types in one go; b) there's no need to parse/stringify anything, as some systems require; c) third-party contributions to that file are validated earlier; and so on.

AnandChowdhary:
We have a deno.land/x mirror that fetches each package from `database.json` here: https://github.com/denorg/x.

It's trivial for me to change the download script to support the `package.ts` file, but my general idea is that JSON is more universal, better-supported in other contexts (non-JavaScript scripts, for example), and better-understood by computers that don't use IntelliSense.

AnandChowdhary:
But I think this might become moot as Deno's package registry evolves and doesn't have a `database.json` anymore: https://github.com/denoland/deno_website2/issues/406

parzhitsky:
Ooo, I’m glad they’re thinking this through! The whole approach with npm-ish global scope of unique names updated manually through pull requests seamed flawed from the start.

Thanks for the link!

lucacasonato:
I think JSON is better suited here and there are very valid usecases for importing this json file externally. I think keeping it as JSON makes more sense because it is more accessible. We have tests to catch style errors in the database anyway.

ry:
I appreciate the patch, but JSON for data seems more versatile to me.

PR #509, opened 2020-05-15
Add support to Open Graph Protocol

### LINKED ISSUE #488: consider adding opengraph cover images
When someone links the Deno website the cover image will be easier to scan and will probably be helpful for sharing or references purposes (in personal wikis for examples that embed a cover)

Of course it is a small detail but it can be a great addition ! 😄 

Current preview
[See on https://metatags.io/](https://metatags.io/)

I'm using Notion and when I create a bookmark of Deno it's like this : 
Notion's bookmark of Deno

Of course it's specific to Notion's behavior, just thought it would be a nice thing to have.

mechamobau:
### In this Pull Request:

🔍 adds support to OpenGraph
🔍 adds robots.txt

### Reference issue
Fixes #488


lucacasonato:
:+1:. Could you make this into a reusable component though so we don't have to duplicate all of these meta tags (like the image) everywhere and instead just use the component in the `<Head>` tag? The description and name would be parameters on this component.

mechamobau:
Good suggestion, but I tried this approach, and the `<meta />` was not loading on the page. So I search in the documentation of Next/head and I found this:
image

The `<meta />` tags need to be direct children of `<Head />`.

https://nextjs.org/docs/api-reference/next/head

pages/benchmarks.tsx
```diff
@@ -46,7 +46,27 @@ const Benchmarks = () => {
     <>
       <Head>
         <title>Benchmarks - Deno</title>
+
+        <meta name="title" content="Benchmarks - Deno" />
         <meta name="description" content="Continuous benchmarks for Deno." />
+
+        <meta name="twitter:card" content="summary_large_image" />
+        <meta name="twitter:url" content="https://deno.land" />
```
ry, line 54: This isn't the URL of the benchmarks page?

ry:
Why are both the twitter and open graph things needed? Does twitter not understand open graph?

lucacasonato:
> The `<meta />` tags need to be direct children of `<Head />`.
> 
> https://nextjs.org/docs/api-reference/next/head

@mechamobau what about something like this?

```js
function graphMeta(a, b) {
  return [
    <meta />,
    <meta />,
    ect...
  ];
}

<Head>
  {...graphMeta(foo, bar)}
</Head>
```

mechamobau:
> > The `<meta />` tags need to be direct children of `<Head />`.
> > https://nextjs.org/docs/api-reference/next/head
> 
> @mechamobau what about something like this?
> 
> ```js
> function graphMeta(a, b) {
>   return [
>     <meta />,
>     <meta />,
>     ect...
>   ];
> }
> 
> <Head>
>   {graphMeta(foo, bar)}
> </Head>
> ```

This is a better approach, I'll make that

pages/benchmarks.tsx
mechamobau: Yes, I'll change that URL to `https://deno.land/benchmarks`

mechamobau:
> Why are both the twitter and open graph things needed? Does twitter not understand open graph?
I'll keep just the Twitter meta tag `twitter:card` (because is required)

kidonng:
Just here to mention, remove `the` in `Fixes the #488` or GitHub won't link this PR and the issue 🙂

mechamobau:
> Just here to mention, remove `the` in `Fixes the #488` or GitHub won't link this PR and the issue 🙂

Tks! 😅

components/Registry.tsx
```diff
@@ -172,7 +173,12 @@ const Registry = () => {
           {name}
           {version && `@${version}`} - deno.land/x
         </title>
-        <meta name="description" content="A third party module for Deno." />
+        {metaDescription({
+          title: "deno.land/x",
+          description: "A third party module for Deno.",
+          url: "https://deno.land/v1/",
```
ry, line 179: I think this component is associated with many URLs, none of which are https://deno.land/v1/

mechamobau: fixed 🙂

PR #512, opened 2020-05-16
Add deno_structured_logging

Yamboy1:


Yamboy1:
Deno Structured Logging (DSL) is an extensible logging system with inspiration from Serilog. It makes it easy to send logs to multiple places and to filter out log levels, as well as support for structured logging.

PR #514, opened 2020-05-16
Better link styling

### LINKED ISSUE #457: ux: "active" link indicator in the sidenav for the manual
A page on the manual

Adding a distinction between an active and a normal link in the sidenav would greatly improve the UX

Because i'm often trying to CTRL+F to find where I was in the sidenav, little details you know !

buckle2000:
image
Now it looks like this. Fixed visibility issues.

(doesn't fix #457)

### DIFF
--- a/components/Header.tsx
+++ b/components/Header.tsx
@@ -55,9 +55,12 @@ const Header = ({ subtitle }: { subtitle?: string }) => {
           </Link>
           <Link href="/[identifier]" as="/manual">
             <a className="ml-10 font-medium text-gray-500 hover:text-gray-900 transition duration-150 ease-in-out">
-              The Manual
+              Manual
             </a>
           </Link>
+          <a href="https://doc.deno.land/https/github.com/denoland/deno/releases/latest/download/lib.deno.d.ts" className="ml-10 font-medium text-gray-500 hover:text-gray-900 transition duration-150 ease-in-out">
+            Runtime API
+          </a>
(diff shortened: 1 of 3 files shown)


lucacasonato:
I don't know about others, but I like the current styling better. Might just be me though.

kennanseno:
> I don't know about others, but I like the current styling better. Might just be me though.

Prefer the current one as well actually, keeping it black feels better overall as it aligns with the overall theme of the website. I think the highlight to blue if it's the active page suffiices.

buckle2000:
I don't know about "look", but the current contrast on hover is not enough. Also, there is no indication of which pages you have visited, it would be hard for users to remember.

buckle2000:
We con try inverted colors too.

lucacasonato:
What about underline on hover or focus in addition to the slight change of color we have right now, and no indication of what pages you visited? Most pages don't give an indication of that anyway.

buckle2000:
underline + color change is ok. I still think visited links should change color to help with remembering which pages you have visited. It is the default style of `<a>`. Wikipedia has it.  Google used to have that feature (now it opens in new tab). Most sites don't have feature that because it is extra code.

ry:
Thanks for the patch, but it must be green to be considered. Due to the high volume, I'm not reviewing red PRs, just closing to clear out the queue. I apologize if I haven't addressed something. Please open a new PR.

PR #519, opened 2020-05-16
Highlight active TOC heading

### LINKED ISSUE #457: ux: "active" link indicator in the sidenav for the manual

### LINKED ISSUE #515: Scroll TOC to current section on page load
For example, the following page (9.1) is not visible in navigation when opened:
image

This should be fixed for mobile too.

kidonng:
Fix #457
Fix #515
Related #514 

- Sync TOC scroll position with current page
- Also made first level headings bold for further distinction.
lDsln5rgE0

### DIFF
--- a/components/Manual.tsx
+++ b/components/Manual.tsx
@@ -179,7 +179,7 @@ function Manual() {
                     />
                   </div>
                   {tableOfContents && (
-                    <ToC tableOfContents={tableOfContents} version={version} />
+                    <ToC tableOfContents={tableOfContents} version={version} path={path} />
                   )}
                 </div>
               </Transition>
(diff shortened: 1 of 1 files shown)


components/Manual.tsx
```diff
@@ -54,6 +54,13 @@ function Manual() {
   useEffect(() => {
     getTableOfContents(version ?? "master")
       .then(setTableOfContents)
+      .then(() =>
+        setTimeout(
```
AnandChowdhary, line 58: Maybe `requestAnimationFrame` instead of `setTimeout` 0?

kidonng: `setTimeout` is used in other places too:

https://github.com/denoland/deno_website2/blob/4506a164821cf4f222bb65fd8d1e92fe36ed7a8a/components/Markdown.tsx#L110-L119

And `requestAnimationFrame` is not a **macrotask** like `setTimeout`. It may not trigger on every event loop.

kidonng:
#515 is not fixed on mobile 😔

PR #521, opened 2020-05-16
Friendlier sidebar

### LINKED ISSUE #494: Auto close side bar on mobile
Clicking on a link in the side bar on mobile devices does not close it, making it more difficult to go through the docs.  
It would feel much more natural if it closed when clicking a different section.

kidonng:
Fix #494 

- Close sidebar when clicking outside
- Close sidebar when navigating away

kaMeFdoagY

### DIFF
--- a/components/Manual.tsx
+++ b/components/Manual.tsx
@@ -43,7 +43,7 @@ function Manual() {
 
   const [showSidebar, setShowSidebar] = useState<boolean>(false);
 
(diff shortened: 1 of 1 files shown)


lucacasonato:
I think the close button should stay, but clicking anywhere else should also close the sidebar. It is not always obvious for everyone that you can click outside of the sidebar to close it.

kidonng:
> I think the close button should stay

Personally I would like it go simple 🙂 but well that's just personal preference. Added back the button.

PR #522, opened 2020-05-16
Use CDN when fetching source files

kidonng:
Currently we are using `raw.githubusercontent.com` to fetch source files **in browsers**. However, some people (e.g. in Mainland China) may have problems accessing the resources. Using a CDN both solves the inconvenience and accelerates browsing.

I suggest using [jsDelivr](https://www.jsdelivr.com/) given it is widely used and works in Mainland China.

This PR:

- Use jsDelivr when **browsers** fetch **manual** files
- Use jsDelivr when **browsers** fetch **registry** files

If you think the change only benefits particular people, we can make it a switch.

### DIFF
--- a/util/manual_utils.test.ts
+++ b/util/manual_utils.test.ts
@@ -13,6 +13,6 @@ test("get introduction file", async () => {
   expect(
     getFileURL("f184332c09c851faac50f598d29ebe4426e05464", "/introduction")
   ).toEqual(
-    "https://cdn.jsdelivr.net/denoland/deno@f184332c09c851faac50f598d29ebe4426e05464/docs/introduction.md"
+    "https://cdn.jsdelivr.net/gh/denoland/deno@f184332c09c851faac50f598d29ebe4426e05464/docs/introduction.md"
   );
 });
--- a/util/manual_utils.test.ts
+++ b/util/manual_utils.test.ts
@@ -13,6 +13,6 @@ test("get introduction file", async () => {
   expect(
     getFileURL("f184332c09c851faac50f598d29ebe4426e05464", "/introduction")
   ).toEqual(
-    "https://raw.githubusercontent.com/denoland/deno/f184332c09c851faac50f598d29ebe4426e05464/docs/introduction.md"
+    "https://cdn.jsdelivr.net/denoland/deno@f184332c09c851faac50f598d29ebe4426e05464/docs/introduction.md"
   );
 });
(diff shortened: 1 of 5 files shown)


lucacasonato:
Are there issues in mainland china regarding reaching GitHub? Because if this is the case it might be worth considering, but it would be great to introduce as little extra hops as possible. If this works we should keep it because it makes it very transparent to the user what is happening, and we are not introducing another party that you need to trust when using `/x`.

kidonng:
> Are there issues in mainland china regarding reaching GitHub?

Part of GitHub is blocked (`gist.github.com`), `*.githubusercontent.com` is generally slow or inaccessible (because GitHub uses AWS which has poor support in Mainland China).

> we are not introducing another party that you need to trust when using /x.

Loading scripts in Deno goes through CloudFlare and should not be worried about. This PR only changes how visitors using browsers fetch raw data from GitHub.

ry:
I'm hesitant for the same reasons that @lucacasonato is. Although github has its issues, it's a single source of truth.

kidonng:
How about making it an option/switch? It can be put on a separate page so there won't be visual noise for most people.
This problem is hard to hack (with User Script, etc.), and making a mirror/fork site for this sole purpose seems futile.

lucacasonato:
@kiddong Easy fix: instead of querying raw source url directly we can just query the deno.land/x url right? Then it will be queried via Cloudflare which has good presence in mainland china afaik

kidonng:
😅 I think you have mistaken my words. I'm stressing that this change only affect  browsers.

> Loading scripts in Deno goes through CloudFlare and should not be worried about. This PR only changes how visitors using browsers fetch raw data (namely manual and registry files) from GitHub.

lucacasonato:
I understand. That’s why I am saying we should have the front end query the Cloudflare worker which queries the source rather than directly querying the source from the front end. 

So this:

HTML deno.land/x/foo/bar.ts -> raw deno.land/x/foo/bar.ts -> raw.githubusercontent.com

instead of

HTML deno.land/x/foo/bar.ts -> raw.githubusercontent.com

kidonng:
This still falls into my statement, it just changes the CDN from jsDelivr (or whatever) to CloudFlare, which is fine.

ry:
@kidonng Thanks for this patch - but we're going to solve this another way - by mirroring github content in s3. This work is being done by @lucacasonato in https://github.com/denoland/deno_registry2

PR #549, opened 2020-05-17
Small tweaks

### LINKED ISSUE #208: render tabs as 4
The default tab size of 8 is excessive for code.
```css
tab-size: 4;
-moz-tab-size: 4;
```

kidonng:
Fix #208 

- Use 2 spaces for tabs
- Reset scroll position (scroll to top) when navigating through manuals


ry:
Thanks.

Do we even have tabs anywhere? They should be size 2.

kidonng:
> Do we even have tabs anywhere?

See example in the issue: https://github.com/denoland/deno_website2/issues/208#issuecomment-629562040

> They should be size 2 if anything.

GitHub uses 4. On second thought, it aligns with most languages that use 4 spaces, but for JavaScript we'd better use 2.

PR #552, opened 2020-05-17
Lossless Image Optimization

FlipFloop:


lucacasonato:
Someone already did this a few days ago, but thanks for the effort :+1:

ry:
Thanks for the patch but I landed a similar change in #744.

PR #557, opened 2020-05-17
Add webmanifest and apple-touch-icon

m00s:
Use the Deno logo when users add the website to the mobile home screen


ry:
What does this do?

m00s:
It correctly shows Deno’s logo and name when you “Add to Home Screen” from a mobile browser.
The apple-touch-icon meta tag is for iOS and the webmanifest is for Android browsers.
I used the icon at `/public/images/logo.svg` to generate the different PNGs

I keep the Deno manual saved on my homescreen and I can never find it because the icon is missing, so I though it was a tiny nice addition.

My original description should have been better, sorry about that

PR #610, opened 2020-05-19
Added pg-format to database.json

grantcarthew:


database.json
```diff
@@ -1323,6 +1323,12 @@
     "repo": "PerlinNoise-Deno",
     "desc": "A simple to use perlin noise module for typescript."
   },
+  "pg-format": {
```
ry, line 1326: no "-" in module name

Issue #617, opened 2020-05-19
github like line highlighted 


rjoydip:

rjoydip:
#620

lucacasonato:
This can be closed

Issue #398, opened 2020-05-13
Should accept url like 'https://deno.land/std/fs/copy.ts:132:8' nor 404


heroboy:
This is what deno output.

kidonng:
Duplicate of #210

PR #620, opened 2020-05-19
Support line numbers and github like line highlighted

rjoydip:
This PR addressed #210, #398 and #617. Implementation is followed GitHub like line references.

//cc @lucacasonato

### DIFF
--- a/components/CodeBlock.tsx
+++ b/components/CodeBlock.tsx
@@ -5,6 +5,7 @@ import Link from "next/link";
 import Router from "next/router";
 import Highlight, { Prism } from "prism-react-renderer";
 import light from "prism-react-renderer/themes/github";
+import { useLayoutEffect } from "react";
 
 export interface CodeBlockProps {
   code: string;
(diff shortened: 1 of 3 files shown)


components/CodeBlock.tsx
```diff
@@ -25,7 +28,36 @@ export const RawCodeBlock = ({
...
+  useEffect(() => {
+    Router.events.on("hashChangeComplete", (url: any) => {
+      setHashValue(url.slice(url.indexOf("#")));
+    });
+    const { hash } = location;
+    setHashValue(hash);
+    return () => {
+      Router.events.off("hashChangeComplete", () => {});
+    };
+  }, []);
```
lucacasonato, line 43: Can't you use the `useRouter` hook for this? Might not be possible, but seems cleaner.

rjoydip: Okay sure

lucacasonato: Oh yeah you're right. Looks good then :-)

lucacasonato:
Can you merge with master to make CI happy?

rjoydip:
//cc @leonardoelias Merged with master

PR #626, opened 2020-05-19
Fixed worker tests

lucacasonato:
This will fail atm, but it will work in master. I need to rewrite the testing for the cf worker to catch this in a pr, before being on master.


lucacasonato:
I guess not needed because of #625

PR #627, opened 2020-05-19

kennanseno:
Not showing anything on dir paths for now rather than 0B which is misleading.

Realistically we should pull total size inside a directory but we've bigger fish to fry.


lucacasonato:
Looks good :+1:. Can you merge with master please?

kennanseno:
@lucacasonato sure thing. closing this one.

PR #659, opened 2020-05-20
feat: virtual database entries for npm and github

pi0:
The first time I came to the deno.land website I was quite surprised not to be able to resolve my repo.

For the listing, of course, it makes sense to make a PR to `database.json` but with this addition, any project can still have an elegant URL while keeping the website scalable. 

New working URLs:

- https://deno.land/x/npm:package
- https://deno.land/x/gh:repo:owner

Preview: 

- https://deno-website2-ei6cxyl1c.now.sh/x/gh:microsoft:typescript
- https://deno-website2-ei6cxyl1c.now.sh/x/npm:destr


lucacasonato:
Ref #406 

I am in favor of this change :+1:. Don't know about @ry.

pi0:
Thanks, @ry for helping to merge. I agree it may not be the best URL pattern for the long term.

srdjan:
nice!

Question (possibly mistyped):

> New working URLs:
> 
> * https://deno.land/x/npm:package
> * https://deno.land/x/gh:repo:owner

shouldn't it be: 
* https://deno.land/x/gh:owner:repo

btw, what is the reason to resolve to the root of the repo?
wouldn't pointing to 'dist' folder, be more proper?

pi0:
@srdjan Yes could be simpler but I think it also adds some opinionated implications in cost of omitting `/dist` from URL that: There is always a build step which should update `dist` and also checked-in into repository. Taking [a1](https://deno.land/x/gh:maheshkareeya:a1) as an example, it is simply using top-level `index.ts`

pi0:
Just for reference seems this solution is deprecated

PR #738, opened 2020-05-23
Fix markdown link parsing in headings

### LINKED ISSUE #562: Link parsing of README is inconsistent with Github Markdown
### Issue
You can see this issue in the README.md file in the status module ([GitHub](https://github.com/denosaurs/status) vs [Deno](https://deno.land/x/status)). The square brackets in the headings are not followed by a `()` but they are still parsed as links in the Deno markdown rendering. This is not the case in the Github markdown rendering of the same README.

### Possible solution
```typescript
// components/Markdown.tsx
const renderers = (canonicalURL: string) => ({
  // ...
  // adding linkReference in the renderers
  linkReference: LinkRenderer,
  // ...
});
```

```typescript
// components/Markdown.tsx
function LinkRenderer(props: LinkRendererProps) {
  // ...
  // adding a check if the href is actually present
  if (href) {
    return (
      <a href={href} className="link">
        {props.children}
      </a>
    );
  }
  return (<>[{props.children}]</>):
}
```

Should I open a pull request to implement these changes?

Qu4k:
Closes #562 

Difference can be seen in the preview deployment:
[`original`](https://deno.land/x/status) vs [`preview`](https://deno-website2-8o3ux5qtk.now.sh/x/status) vs [`GitHub`](https://github.com/denosaurs/status)


ry:
This is a bug that needs to be fixed upstream. By adding this we would not be following commonmark spec anymore. You're is trying to make ‘[foo]’ render like ‘[foo]’ instead of a link ‘foo’ which is what common mark says it should.

Thanks for the patch @Qu4k but closing without merge.

PR #741, opened 2020-05-24
Add Commands -Update database.json

buttercubz:


lucacasonato:
The database entries should be sorted alphabetically. Please insert your entry in the correct location.

PR #746, opened 2020-05-24
add library convert_size

AliBasicCoder:


AliBasicCoder:
why? keys Must be sorted alphabetically? why not sort it in the website instead?

lucacasonato:
Because it's faster to do it once than to have every client sort through a very long hashmap. `User experience` > `Developer experience`

PR #766, opened 2020-05-24
Add iam_policies module

roggervalf:
I try to add my own simple module to database list.
detail standalone repo https://github.com/roggervalf/iam-policies/tree/master/dist


lucacasonato:
The key of your module should not use dashes (`-`) in its name. You should use underscores (`_`) instead.

roggervalf:
@lucacasonato change done already

roggervalf:
@lucacasonato change already done, please let me know if there is something else to be changed

PR #771, opened 2020-05-24
Update database.json

aaronwlee:


lucacasonato:
The key of your module should not use dashes (`-`) in its name. You should use underscores (`_`) instead.

lucacasonato:
@aaronwlee You can still get this merged if you update the key to be `attain_graphqL`. The repository name can stay as is.

PR #786, opened 2020-05-25
Updated .gitignore for ignoring vscode settings folder

Swap76:
I noticed this issue while working on another fix but I think we need to take the standard `.gitignore` file from the create-react-app.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,6 @@ worker/script.js
 
 .now
 .vercel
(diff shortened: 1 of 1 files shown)


nayeemrmn:
It's probably fine since `.DS_STORE` is there, but honestly people should put things like this in their machine-global `.gitignore`s or `.git/info/exclude`.

PR #796, opened 2020-05-25

kennanseno:
https://github.com/denoland/deno_website2/pull/627

@lucacasonato


components/Registry.tsx
```diff
@@ -519,9 +519,14 @@ function DirectoryListing(props: {
...
-                          {entry.size !== undefined
-                            ? bytesToSize(entry.size)
-                            : "N/A"}
+                          {
+                            /** TODO(kennanseno): Directory should pull total size inside it.  */
+                            entry.type === "dir"
+                              ? ""
+                              : entry.size !== undefined
+                              ? bytesToSize(entry.size)
+                              : "N/A"
```
ry, line 528: How about just blank instead of "N/A" ?

kennanseno: empty string could just work to be honest but it's probably better to differentiate between a valid directory and a file for that some reason we can't resolve the size.

ry: > a file that for some reason we can't resolve the size

when does that happen?

kennanseno: I'm just assuming based on the existing undefined check to entry.size.

Not a big deal tbh and happy to leave it to empty string either way.

Swap76:
@kennanseno Thanks :-)

PR #799, opened 2020-05-25
fix(#798): Async callback was not invoked within the 5000ms timeout

rjoydip:
Closes: #798


lucacasonato:
Which test is this for? Many tests explicitly set 5000ms timeouts.

rjoydip:
NPM directory listing

https://github.com/denoland/deno_website2/blob/ba67b17b653aaf5bfaeb485586c3cb3c52f481c5/util/registries/npm.test.ts#L67

PR #801, opened 2020-05-25
fix: purge css warning

rjoydip:
After remove purge css warning build has become faster than before.

### DIFF
--- a/tailwind.config.js
+++ b/tailwind.config.js
@@ -26,4 +26,8 @@ module.exports = {
     backgroundColor: ["responsive", "hover", "focus", "active"],
   },
   plugins: [require("@tailwindcss/ui"), require("tailwindcss-hyphens")],
+  purge: {
+    enabled: true,
+    content: ['./components/**/*.tsx', './pages/**/*.tsx'],
+  },
 };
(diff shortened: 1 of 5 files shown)


lucacasonato:
But then you should also remove the existing purgecss config in the postcss config.

rjoydip:
@lucacasonato Done. removed the existing purgecss configuration from the in the postcss config.

tailwind.config.js
```diff
@@ -26,4 +26,8 @@ module.exports = {
     backgroundColor: ["responsive", "hover", "focus", "active"],
   },
   plugins: [require("@tailwindcss/ui"), require("tailwindcss-hyphens")],
+  purge: {
+    enabled: true,
+    content: ['./components/**/*.tsx', './pages/**/*.tsx'],
+  },
```
lucacasonato, line 32:
```suggestion
  purge: ['./components/**/*.tsx', './pages/**/*.tsx'],
```
The `enabled: true` forcefully enables purging, not just during production (https://tailwindcss.com/docs/controlling-file-size#enabling-manually). We only want to use it for production though, because it makes the dev loop slower to purge on every build.

rjoydip: Resolved on this commit https://github.com/denoland/deno_website2/pull/801/commits/4f316af543b1bbcd04ae451c6938e72c8ac5daa5

PR #830, opened 2020-05-26
Add tweetnacl-deno to database

dr-useless:
Add port of tweetnacl cryptographic library to third-party modules


roggervalf:
Hi @dr-useless, the key of your module should not use dashes (-) in its name. You should use underscores (_) instead. #766

Issue #797, opened 2020-05-25
'View on GitHub' seems out of place


kennanseno:
'View on GitHub' link for the markdowns in the Manual page seems out of place and serves no purpose as user has already read the doc. 

Realistically, we should keep it similar and use the icon similar to what we have in the homepage. Would probably suit better if it's the GitHub icon and just on the same light as the title but on the right side.

<img width="1112" alt="Screenshot 2020-05-25 at 19 30 42" src="https://user-images.githubusercontent.com/7852866/82836448-642f0780-9ebe-11ea-8cf9-f8aafcdf2721.png">

kidonng:
> serves no purpose as user has already read the doc.

https://github.com/denoland/deno_website2/issues/516#issuecomment-629638914

And someone may prefer to read it on GitHub or whatever. If you look around many docs provide link to GitHub as well.

> Realistically, we should keep it similar and use the icon similar to what we have in the homepage.

Good idea.

lucacasonato:
It might be better if it said 'Edit on GitHub' and linked you to the file edit screen on GitHub for that page.

kidonng:
> It might be better if it said 'Edit on GitHub' and linked you to the file edit screen on GitHub for that page.

From my personal perspective, it's not a very good experience. GitHub will automatically fork the repo if you want to make edits, but more often I just want the link to the document.

smkamranqadri:
I would like to take this issue once finalised, in my opinion, edit on github make more sense.

![Screenshot 2020-05-26 at 5 43 03 PM](https://user-images.githubusercontent.com/13115396/82902081-5b9b0780-9f78-11ea-98b8-7447976abbba.png)

PR #831, opened 2020-05-26
change view on github to edit.

smkamranqadri:
As per discussion on #797


ry:
Thanks but I think it's better to link to the file rather than the edit form. Maybe we can move that link somewhere to make it less confusing.

Issue #848, opened 2020-05-27
API Limit exceeds for star badges in third party modules


divy-work:
The start badges that were recently added to the website requests `img.shields.io` that have a API limit:
`Exceeded limit 300 requests per 200 seconds`

So some images appear broken:
![Screenshot from 2020-05-27 12-15-22](https://user-images.githubusercontent.com/34997667/82986936-60ad9480-a014-11ea-82d4-f02225aaa6eb.png)

Swap76:
@divy-work Can you provide the specifications of the browser as on the third-party modules page for me all badges are visible.

lucacasonato:
@divy-work I am getting this too. I think we should revert #603 for now because of the rate limit. I see more not loaded badges than loaded badges: 
![image](https://user-images.githubusercontent.com/7829205/83007436-f7c61c80-a013-11ea-8949-7bfec63e1b53.png)

PR #603, opened 2020-05-18
feat: added stars for github and download badges for npm

rjoydip:
This PR is created for #464 

`star badge` for `github` and `download badge` for `npm` project. Due to enable this feature `utility` modules has been refactored. 

***Star Badge***

image

***Download  Badge***

image

***Repo not found***

image

PR #852, opened 2020-05-27
Revert "feat: added stars for github and download badges for npm"

lucacasonato:
Reverts denoland/deno_website2#603

A lot of the badges are not actually displayed because of rate limits:
image

Also this is probably causing a lot of strain on `img.shields.io` because we request ~400 badges for every /x load. We should come up with a different solution that includes caching on our side.

Ref #848

### DIFF
--- a/pages/x/index.tsx
+++ b/pages/x/index.tsx
@@ -128,27 +128,6 @@ const ThirdPartyRegistryList = () => {
                                 </div>
                               </div>
                               <div>
-                                {entries[name]?.type === "github" ? (
-                                  <img
-                                    src={`https://img.shields.io/github/stars/${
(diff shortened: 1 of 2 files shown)


Swap76:
@lucacasonato Maybe in future updated we need pagination which will only load 50 or 100 a page & also the download badge for npm packages was weird (out of sync from stars)

lucacasonato:
I agree with the pagination for the future, but still it is also pretty irresponsible of us to request so many badges from shields.io - it is community run AFAIK. I can see spikes in resource use whenever I load /x: https://metrics.shields.io/d/g_1B7zhik/prom-client-default-metrics :sob:

PR #875, opened 2020-05-28
Fix style code: Prettier to correct the CI error

mtwzim:
I saw that the recent pull's request is wrong with the CI in relation to Prettier, here's a correction.


ry:
corrected in #659. Thanks.

PR #878, opened 2020-05-28
Added boolbase in Third Party Modules GitHub

mtwzim:


fransafu:
Hi @ry!

In a group of programmers, we commented about this module and we don't have an idea what is its purpose. Maybe I can't understand It, in the case if which was transcribed module of JS to Typescript without reason or the main motive has only put a "commit" in Deno project... I recommend removing this module.

fransafu:
This project is based on https://github.com/fb55/boolbase and it doesn't have an update for a long time.

I suggest checking contributions of the user @mtwzim

PR #879, opened 2020-05-28
Added nth-check in Third Party Modules GitHub

mtwzim:


mtwzim:
This error was not caused by my change.

alirealasad:
"-" doesn't support by deno so change the name to nthCheck, nth_check or something else

mtwzim:
> "-" doesn't support by deno so change the name to nthCheck, nth_check or something else

Solved

PR #940, opened 2020-05-30
Add maze_generator to third party modules

TheWizardBear:
https://github.com/TheWizardBear/maze_generator


TheWizardBear:
The test says that it probably failed because I forgot to run Prettier, but I'm not exactly sure how to do that?

PR #954, opened 2020-05-31
Add deno-aws-sign-v4 back to database.json

ccchapman:
This was removed by #837 after being added in #689. It appears like the repository owned had changed their GitHub username, or perhaps it was added incorrectly in the first place.

Refs:

cde8674 Removed repos which does not exists (#837)
a5fd875 Adding deno-aws-sign-v4 (#689)

cc @ry @Swap76 @silver-xu

### DIFF
--- a/database.json
+++ b/database.json
@@ -633,6 +633,12 @@
     "repo": "deno_app_setuper",
     "desc": "A CLI to setup apps like a Create React App."
   },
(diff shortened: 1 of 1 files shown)


Swap76:
@ccchapman LGTM we have deleted the package from database as this was no longer visible at that time but now it's perfect 🙌

Issue #952, opened 2020-05-31
Delete uuid in third party modules


buttercubz:
i think uuid should be removed from third party modules as it can create confusion, and maybe third party modules should not have std module names.

PR #955, opened 2020-05-31
Remove third party module lucascaro/deno-uuid

ccchapman:
See standard library package https://deno.land/std/uuid

Closes https://github.com/denoland/deno_website2/issues/952


ry:
That user can remove it if they want

PR #958, opened 2020-05-31
Add web-name to third party list

vimalverma558:
Simple module to generate random usernames

### DIFF
--- a/database.json
+++ b/database.json
@@ -3079,7 +3079,6 @@
     "repo": "deno-watch",
     "desc": "A pure deno file watcher."
   },
-  },
     "web-name": {
     "type": "github",
     "owner": "letskhabar",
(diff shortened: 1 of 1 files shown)


database.json
```diff
@@ -3079,6 +3079,12 @@
     "repo": "deno-watch",
     "desc": "A pure deno file watcher."
   },
+  "web-name": {
+    "type": "github",
+    "owner": "letskhabar",
+    "repo": "web-name",
+    "desc": "simple module to generate random usernames"
+  },
```
alirealasad, line 3087: database names with dashes are not allowed so can change the name to web_name, webName something else

vimalverma558: Thanks for suggestions 😊😘

s1mpson: > database names with dashes are not allowed so can change the name to web_name, webName something else

@alirealasad Are module names case-sensitive? That would be a disaster having different modules named `webname`, `webName`, `Webname`, `wEbname` and so on

vimalverma558: Web_name is also a option

PR #974, opened 2020-06-01
Add 0vw/yo_mamma to the database

0vw:
yo_mamma is https://github.com/Fadyazmy/yo-mamma ported to Deno


narwy:
How does this benefit anyone?

0vw:
> How does this benefit anyone?

Well it's a module I use a lot in node. It's quite fun for stuff such as discord bots.

PR #980, opened 2020-06-01
Add color-deno module

selvakumardhivakar:
Vuelancer created color-deno module for printing colored log in console for deno

### DIFF
--- a/database.json
+++ b/database.json
@@ -384,17 +384,17 @@
     "repo": "color",
     "desc": "Write Colourful Messages In console. It's easy to learn and use. ༼ つ ◕_◕ ༽つ"
   },
-  "colors": {
-    "type": "github",
-    "owner": "maheshkareeya",
-    "repo": "colors",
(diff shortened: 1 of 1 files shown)


AvnerCohen:
Deno has colors in the std lib:  https://github.com/denoland/deno/blob/master/std/fmt/colors.ts

selvakumardhivakar:
**colorlog** using only green, red, yellow colors for indicating success, error and warning.

selvakumardhivakar:
> Deno has colors in the std lib: https://github.com/denoland/deno/blob/master/std/fmt/colors.ts

Now only I noticed the standard library for colors in deno. 
> I think this _colorlog_ will be used only for indicating success, error and warning.

rjoydip:
@selvakumardhivakar `colorlog` looks like a very simple utility it could be achieve through deno colors std modules. But if you could extends `colorlog` functionality more like [signale](https://github.com/klaussinani/signale) modules do in node.

PR #989, opened 2020-06-01
add 0vw/teejay to the database

0vw:
Returns a random phrase from the popular author and philosopher known as Teejayx6


0vw:
Messed up order.

PR #1030, opened 2020-06-03
add makehelp

konsumer:
This adds [`makehelp`](https://github.com/konsumer/deno-make-help) 3rd-party module, which is a very simple script to parse your Makefile, and output help. I think this will be useful for people wishing to replace `npm run` with plain Makefiles.


konsumer:
I made the tests pass, on CI, but I can't get the tests to run locally, because I get a bunch of unrelated errors.

PR #1061, opened 2020-06-06
Add Houston mod to database.json

MartonDev:


database.json
```diff
@@ -1668,6 +1668,12 @@
     "repo": "hmac",
     "desc": "Convenience function for macing singular data. hash should be one of sha1, sha256, or sha512"
   },
+  "Houston": {
```
alirealasad, line 1671: change the name to houston to fix the ci because uppercases are  not allowed

PR #1066, opened 2020-06-06
Remove meta twitter card stuff

ry:
Too complicated and results in links with bad titles.


mechamobau:
Can you give an explanation about why it is too complicated and needed to be removed?

PR #1067, opened 2020-06-06
added string-builder module.

bqio:


database.json
```diff
@@ -3180,6 +3180,12 @@
     "repo": "strcolors.ts",
     "desc": "FMT String Extensions for Deno"
   },
+  "string-builder": {
```
alirealasad, line 3183: change the name to stringBuilder, string_builder or something else

PR #1096, opened 2020-06-10
Change random thirdparty module filtering

littletof:
The current filter only chooses modules, which description's length is between 70 and 90. I assume it is to make sure they have a 3 line description on the page, so truncation doesnt cause a problem

I checked it, and currently only 131 entries satisfy this, out of 629 (20%).
As 12 modules are shown on the page on each load, the modules shown repeat quite often.

I tought it might be good to have all modules in the pool.
Of course with this there can be too long desctiptions, so i truncated them using `-webkit-line-clamp`. [caniuse](https://caniuse.com/#search=-webkit-line-clamp) tells its well supported.
I tried Chrome, Firefox, Edge, IE. All work except IE of course (but the current page has problems too)

I understand this may be a design and UI decision, so feel free to close this PR.

### DIFF
--- a/pages/index.tsx
+++ b/pages/index.tsx
@@ -35,13 +35,9 @@ for await (const req of s) {
     SimpleEntry[] | null
   >(null);
   useEffect(() => {
-    const thirdPartySelection = [];
-    for (let i = 0; i < NUM_THIRD_PARTY; i++) {
-      const s = Math.floor(thirdPartyEntryPool.length * Math.random());
-      thirdPartySelection.push(thirdPartyEntryPool[s]);
-      thirdPartyEntryPool.splice(s, 1);
-    }
-    setThirdPartySelection(thirdPartySelection);
+    setThirdPartySelection(
+      RandomEntriesFromArray(thirdPartyEntryPool, NUM_THIRD_PARTY)
+    );
   }, []);
 
   return (
@@ -266,7 +262,13 @@ for await (const req of s) {
                 <h4 className="text-lg font-bold">{s.name}</h4>
                 <p
                   className="whitespace-normal break-words text-gray-700 mt-2 overflow-hidden"
-                  style={{ textOverflow: "ellipsis", minHeight: "4.5em", WebkitLineClamp: 3, WebkitBoxOrient: "vertical", display: "-webkit-box" }}
+                  style={{
+                    textOverflow: "ellipsis",
+                    minHeight: "4.5em",
+                    WebkitLineClamp: 3,
+                    WebkitBoxOrient: "vertical",
+                    display: "-webkit-box",
+                  }}
                 >
                   {s.desc}
                 </p>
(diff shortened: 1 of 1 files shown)


lucacasonato:
@littletof I'm generally in favor, but adding these extra entries slows initial page load because they are all inlined into the static html file that gets generated by Next.js. Just the batch of extra modules in this PR add 29KB of extra data on first load. That is not great. Especially because more and more modules are added every day.

I think the easiest fix would be make a random selection of ~48 modules during static build, which can then further be selected from in the frontend. As the page gets deployed multiple times every day the random selection of 48 should change multiple times a day. That we can show more modules over the course of a day, while keeping page load times under control. What do you think?

littletof:
@lucacasonato well, if it brings down performance, then the selected pool on each deploy, like you proposed, could work instead, while there are multiple deploys a day. The main point is to not restrict the referencing of modules to be based only on their descriptions length.

I will give it a go, but I'm not familiar with React, so feel free, to implement it in an other branch.

pages/index.tsx
```diff
@@ -386,8 +386,15 @@ export const getStaticProps: GetStaticProps<HomeProps> = async () => {
     }
   });
 
+  const thirdPartyEntryPool = [];
+  for (let i = 0; i < POOL_NUM_THIRD_PARTY; i++) {
+    const s = Math.floor(thirdPartyEntries.length * Math.random());
+    thirdPartyEntryPool.push(thirdPartyEntries[s]);
+    thirdPartyEntries.splice(s, 1);
+  }
```
littletof, line 394: This whole random selection along with the one in the useEffect should be moved to a utility method. Where should I put it? Is putting it to `manual_utils` or `register_utils` acceptable?

pages/index.tsx
```diff
@@ -264,8 +265,8 @@ for await (const req of s) {
               >
                 <h4 className="text-lg font-bold">{s.name}</h4>
                 <p
-                  className="whitespace-normal break-words text-gray-700 mt-2"
-                  style={{ textOverflow: "ellipsis" }}
+                  className="whitespace-normal break-words text-gray-700 mt-2 overflow-hidden"
+                  style={{ textOverflow: "ellipsis", minHeight: "4.5em", WebkitLineClamp: 3, WebkitBoxOrient: "vertical", display: "-webkit-box" }}
```
littletof, line 269: I'm having trouble with prettier not wanting to format, hence the linter gives an error.

pages/index.tsx
lucacasonato: I think in `pages/index.ts` is probably fine as that is the only place its used :-)

pages/index.tsx
lucacasonato: Did you run `yarn format`?

littletof: I ran it by hand, because I wanted to avoid installing yarn just for this (I use npm :) ) I now noticed I missed the write flag. It writes all the files, but I can stage only the one I modified so this is sorted now

pages/index.tsx
```diff
@@ -367,15 +370,27 @@ const InstallSection = () => {
   );
 };
 
+const RandomEntriesFromArray = (sourceArray: SimpleEntry[], count: number) => {
+  const source = sourceArray;
```
littletof, line 374: Wanted to copy this array like `[...sourceArray]`, but it gave me some error with `Circular references cannot be expressed in JSON`, so there is some React or next mechanism that I clearly dont understand, but the selection seems to work fine without the copy, so I took it out.

littletof:
@lucacasonato it works now, after your review its done I think.

littletof:
@lucacasonato thanks!

Using ellipses would feel a bit odd for me, from a UI/UX standpoint, but dont really have a better solution other then these two you mentioned.

PR #1117, opened 2020-06-13
Add Black Lives Matter banner to top of site

lucacasonato:


ry:
I don't want to make empty gestures of [performative allyship](https://forge.medium.com/performative-allyship-is-deadly-c900645d9f1f). The solution to racism isn't that we all wear anti-racism armbands. I encourage anyone concerned about police violence - which should be everyone - to look into the organizations who protect bad cops; in particular to understand [what police unions in the US do](https://www.nytimes.com/2020/06/06/us/police-unions-minneapolis-kroll.html) and [which politicians they support](https://www.npr.org/2016/09/18/494451660/the-nations-largest-police-union-endorses-donald-trump). I hope people, especially Americans, will vote for policy changes that actually make a difference.

piscisaureus:
@lucacasonato I absolutely believe that you submitted this PR with honestly good intentions.

However, as one of the Deno maintainers I don't feel it would be genuine to put this message up at this time. It's been 3 weeks since the killing of George Floyd and I think everybody's heard about it by now. Putting a banner on our website isn't going to make a difference now when it comes to the actual issue that is systemic racism. To me it'd feel more like a half hearted attempt to "prove" that I am one of the good people, and/or an attempt to keep activist mobs away, and I'm not actually interested in doing that.

For us at this point, meaningfully contributing means addressing social injustice in our own community, and to consider and address the things we are personally doing wrong (surely, unintentionally) in our dealing with minorities .

Issue #1095, opened 2020-06-10
module's default branch different than master


oganexon:
> If you leave out the branch, it will default to the module’s default branch, usually `master`.

This is the case when importing a module but not on x/
_Edit : importing module also doesn't work when `master` doesn't exist_. 

For example, on my github my main branch is `latest` and there is no `master` branch.
The site should take the main branch and not `master`.

https://deno.land/x/remove_forever@v1.0.0 works
https://deno.land/x/remove_forever don't 

Edit : I created the master branch so my example isn't relevant anymore but the issue is still there

lucacasonato:
Use the default_version option in database.json to choose which branch the module should default to.

oganexon:
I'm gonna stay in the master branch , so it's just cosmetics.

PR #1132, opened 2020-06-15
Require explicit default_version

### LINKED ISSUE #1130: Do not assume `master` is the main branch in GitHub
I moved oak's main branch from `master` to `main` and now but the website redirects `https://deno.land/x/oak/mod.ts` to `https://github.com/oakserver/oak/blob/master/mod.ts` instead of `https://github.com/oakserver/oak/blob/main/mod.ts`.

lucacasonato:
Closes #1130

### DIFF
--- a/util/registries/github.ts
+++ b/util/registries/github.ts
@@ -6,16 +6,16 @@ export interface GithubDatabaseEntry extends DatabaseEntry {
   type: "github";
   owner: string;
   repo: string;
+  default_version: string;
   path?: string;
-  default_version?: string;
 }
 
 export class GithubEntry implements Entry {
   public desc: string;
   private owner: string;
   private repo: string;
+  private defaultVersion: string;
   private path?: string;
-  private defaultVersion?: string;
 
   constructor(databaseEntry: GithubDatabaseEntry) {
     this.desc = databaseEntry.desc;
@@ -27,12 +27,12 @@ export class GithubEntry implements Entry {
 
   getSourceURL(path: string, version?: string): string {
     return `https://raw.githubusercontent.com/${this.owner}/${this.repo}/${
-      version ?? this.defaultVersion ?? "master"
+      version ?? this.defaultVersion
     }${this.path ?? ""}${path}`;
   }
   getRepositoryURL(path: string, version?: string): string {
     return `https://github.com/${this.owner}/${this.repo}/tree/${
-      version ?? this.defaultVersion ?? "master"
+      version ?? this.defaultVersion
     }${this.path ?? ""}${path}`;
   }
   async getDirectoryListing(
@@ -43,7 +43,7 @@ export class GithubEntry implements Entry {
       const url = `https://api.github.com/repos/${this.owner}/${
         this.repo
       }/contents${this.path ?? ""}${path}?ref=${
-        version ?? this.defaultVersion ?? "master"
+        version ?? this.defaultVersion
       }`;
       const res = await fetch(url, {
         headers: {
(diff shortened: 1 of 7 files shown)


util/registries/github.ts
```diff
@@ -6,16 +6,16 @@ export interface GithubDatabaseEntry extends DatabaseEntry {
   type: "github";
   owner: string;
   repo: string;
+  default_version: string;
   path?: string;
-  default_version?: string;
```
ry, line 10: default_branch?

ry: Why does every entry have to specify it? Can't we assume master as default ?

lucacasonato: Can be. It's called `default_version` though because version is more universal than branch. (because `"type": "url"` also has a `default_version`?)

ry: Sorry - I thought this patch was adding the "default_version" feature - but it seems that already exists. This is making it required. Why is that desirable? It's more data in the database.json

lucacasonato: > Why does every entry have to specify it? Can't we assume master as default ?

That's what this branch PR changes. `default_version` is already supported on currently deployed site. The reason I wanted to remove the default being `master` is that people assume that we pull the default branch info from GitHub, so are confused when they don't use master as a default branch and everything breaks. Ref #1095 and #1130

ry: I don't particularly care about this - I'll approve it if you insist - it seems some documentation about the "default_version" in the readme would also solve the problem.

lucacasonato: I think most people don't read the README. Also there is a move towards not using `master` as the default git branch, so that kind of aligns with that too, because it makes us not opinionated to `master` being the main branch. I think explicit is better than implicit. But if you think its not worth it I can add docs to readme / pr template and close this.

jozsefsallai:
I agree with @ry here. We could assume that the default branch is `master`, since that's the most common branch name in git, despite the current controversies surrounding the word. Just specifying the `default_version` directly should be enough if your default branch is called something else, otherwise there is no harm in defaulting to `master`.

**But what if `main`, `principal`, etc. will be the most commonly used names in future Deno modules, especially with Github's plans regarding this subject?**

It's still a better idea to specify a default than to add an extra line to every single module that exists in the database.

Alternatively, we could just get the `default_branch` property from the API: https://developer.github.com/v3/repos/#get-a-repository but that would require an extra request.

---

I dislike the idea of mixing politics with software development. It's not worth giving up optimization/performance and backwards-compatibility just to adapt to a trend that was unnecessary in the current context to begin with.

nayeemrmn:
@jozsefsallai You're making presumptions regarding the motivation behind this patch. Read https://github.com/denoland/deno_website2/pull/1132#discussion_r440144180, these are practical concerns based on the actions others are taking.

> I dislike the idea of mixing politics with software development.

Only you have done this yet.

jozsefsallai:
@nayeemrmn The bottom part was just an addendum, it didn't have direct relation to the patch or my comment as a whole (which is why I added a divider too). It's just my view on this whole new trend of shifting away from `master`. I agree it was kinda off-topicy, but when you look at _why_ these shifts happen, it does make a little sense.

I understand the concern, but I just don't see the point of **not** having a fallback (also, I think `default_version` should be better documented, to cover cases like this too). Nobody stops you from specifying `default_version`. Having a fallback doesn't hurt anybody. That's just my 2c.

lucacasonato:
We can not use the GitHub API because of rate limits. Not for stars, not for forks and not for default branches.

PR #1149, opened 2020-06-16
Update database.json to include the module isEven

ajimae:
- a deno module to check if a given number is an even number


hayd:
It'd be reasonable to have a library with a suite of `is` functions, but this (exporting a single one line function) is absurd.

ry:
I agree with @hady

Issue #1163, opened 2020-06-18
highlight style issue in toml file


rjoydip:
Found styling issue (`token punctuation`) in `.toml` file. I think the issue is mainly with table class of  `tailwindcss`.

### Currently
![image](https://user-images.githubusercontent.com/15318294/85019981-45562500-b18d-11ea-926f-fe93b794e976.png)

### Expected
![image](https://user-images.githubusercontent.com/15318294/85019893-28215680-b18d-11ea-98ce-69eb43006380.png)

PR #1164, opened 2020-06-18
fix: style issue in toml file

rjoydip:
PR for #1163 

Currently: https://deno.land/x/tar/Cargo.toml#L1
Actually: https://deno-website2-19p6dftxh.vercel.app/x/tar/Cargo.toml#L1

### DIFF
--- a/components/app.css
+++ b/components/app.css
@@ -16,6 +16,10 @@ html {
   outline: none;
 }
 
+.table {
+  display: inline;
+}
+
 .link {
   @apply text-blue-500;
   @apply transition;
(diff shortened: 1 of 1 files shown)


components/app.css
```diff
@@ -16,6 +16,10 @@ html {
   outline: none;
 }
 
+.table {
+  display: inline;
+}
+
```
lucacasonato, line 22: Now all tables are `display: inline;`. That is not good. This should only be the class is used in a code block.

rjoydip: Now it only targeting table for toml lang [68df5R20](https://github.com/denoland/deno_website2/pull/1164/files#diff-243e12fe4a5aeeaa414c79fdfb968df5R20)

Issue #1170, opened 2020-06-19
Add support for word searching


sandeshdamkondwar:
<img width="1920" alt="Screenshot 2020-06-19 at 2 52 26 PM" src="https://user-images.githubusercontent.com/2510560/85117830-17c8b480-b23d-11ea-98b8-0aacb1339ae9.png">
<img width="1920" alt="Screenshot 2020-06-19 at 2 52 12 PM" src="https://user-images.githubusercontent.com/2510560/85117835-18f9e180-b23d-11ea-90ae-61b2804f2310.png">
<img width="1920" alt="Screenshot 2020-06-19 at 2 46 26 PM" src="https://user-images.githubusercontent.com/2510560/85117838-1a2b0e80-b23d-11ea-94f9-f5db6bc5c054.png">
<img width="1920" alt="Screenshot 2020-06-19 at 2 53 07 PM" src="https://user-images.githubusercontent.com/2510560/85117814-113a3d00-b23d-11ea-9e19-997d2e87e00f.png">

lucacasonato:
Implemented in #1171

PR #1171, opened 2020-06-19
feat: added support fuzzy word searching for X party modules

sandeshdamkondwar:
Added support for fuzzy word searching in X party modules list. 

#1170
Screenshot 2020-06-19 at 2 52 26 PM
Screenshot 2020-06-19 at 2 52 12 PM
Screenshot 2020-06-19 at 2 46 26 PM
Screenshot 2020-06-19 at 2 53 07 PM

### DIFF
--- a/pages/x/index.tsx
+++ b/pages/x/index.tsx
@@ -21,13 +21,7 @@ const ThirdPartyRegistryList = () => {
   const filteredList = useMemo(() => {
     const queryWords = query.split(" ").filter((word) => word.length); // Split by space & ignore empty ones
     const nameReg = new RegExp(queryWords.join("|"), "i"); // Match any word of search query
-    const descReg = new RegExp(
-      queryWords.reduce((a, b) => {
-        if (a === "") return b;
-        return a + "[^" + b + "]*" + b;
-      }, ""),
-      "i"
-    );
+    const descReg = new RegExp(queryWords.join(".*?"), "i");
 
     return Object.keys(entries).filter((name) => {
       const desc = entries[name].desc;
--- a/pages/x/index.tsx
+++ b/pages/x/index.tsx
@@ -19,16 +19,20 @@ const ThirdPartyRegistryList = () => {
 
   // Sorting the list only if entries are changed
   const list = useMemo(() => {
-    return Object.keys(entries)
-      .sort((nameA, nameB) => nameA.localeCompare(nameB));
+    return Object.keys(entries).sort((nameA, nameB) =>
+      nameA.localeCompare(nameB)
+    );
   }, [entries]);
 
   // Filtering the memoized list
-  const filteredList = useMemo(() => list.filter((name) => {
-      const queryReg = new RegExp(query.split(' ').join('.*?'), 'i');
-      const desc = entries[name].desc;
-      return name.match(queryReg) || (desc ?? "").match(queryReg);
-    }), [list, query]
+  const filteredList = useMemo(
+    () =>
+      list.filter((name) => {
+        const queryReg = new RegExp(query.split(" ").join(".*?"), "i");
+        const desc = entries[name].desc;
+        return name.match(queryReg) || (desc ?? "").match(queryReg);
+      }),
+    [list, query]
   );
 
   return (
--- a/pages/x/index.tsx
+++ b/pages/x/index.tsx
@@ -17,18 +17,18 @@ const ThirdPartyRegistryList = () => {
     setQuery(event.target.value);
   }
 
-  const list = useMemo(
-    () =>
-      Object.keys(entries)
-        .filter(
-          (name) =>
-            name.toLowerCase().includes(query.toLowerCase()) ||
-            (entries[name].desc ?? "")
-              .toLowerCase()
-              .includes(query.toLowerCase())
-        )
-        .sort((nameA, nameB) => nameA.localeCompare(nameB)),
-    [entries, query]
+  // Sorting the list only if entries are changed
+  const list = useMemo(() => {
+    return Object.keys(entries)
+      .sort((nameA, nameB) => nameA.localeCompare(nameB));
+  }, [entries]);
+
+  // Filtering the memoized list
+  const filteredList = useMemo(() => list.filter((name) => {
+      const queryReg = new RegExp(query.split(' ').join('.*?'), 'i');
+      const desc = entries[name].desc;
+      return name.match(queryReg) || (desc ?? "").match(queryReg);
+    }), [list, query]
   );
 
   return (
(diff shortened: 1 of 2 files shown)


pages/x/index.tsx
```diff
@@ -17,18 +17,22 @@ const ThirdPartyRegistryList = () => {
     setQuery(event.target.value);
   }
 
-  const list = useMemo(
+  // Sorting the list only if entries are changed
+  const list = useMemo(() => {
```
rjoydip, line 21: You don't need to use `filter` separately. You could do something like below. 

```ts
const list = useMemo(
    () =>
      Object.keys(entries)
        .filter((name) => {
          const queryReg = new RegExp(
            query.split(" ").reduce((a, b) => a + ".*" + b)
          );
          return (
            name.match(queryReg) || (entries[name].desc ?? "").match(queryReg)
          );
        })
        .sort((nameA, nameB) => nameA.localeCompare(nameB)),
    [entries, query]
  )
```

pages/x/index.tsx
```diff
@@ -17,18 +17,22 @@ const ThirdPartyRegistryList = () => {
...
-          (name) =>
-            name.toLowerCase().includes(query.toLowerCase()) ||
-            (entries[name].desc ?? "")
-              .toLowerCase()
-              .includes(query.toLowerCase())
-        )
-        .sort((nameA, nameB) => nameA.localeCompare(nameB)),
-    [entries, query]
+      list.filter((name) => {
+        const queryReg = new RegExp(query.split(" ").join(".*?"), "i");
```
rjoydip, line 31: > The regex itself can be optimised: .* initially grabs as much as possible, then retracts only if the rest doesn't match. This takes a larger than neccessary amount of backtracking to find out. Instead, you could use a reluctant quantifier: .*?. This attempts to match as little as possible. If a match is found, it is found without backtracking. However, it still does a lot of backtracking if there is no match. We can do better: a[^b]*b (at this point, it doesn't really matter if the quantifier is greedy). A possessive quantifier (a[^b]*+b) would be even better, but javascript doesn't support these. Using a character class deprives us of a join, but see the next point.

For more details [stackexchange](https://codereview.stackexchange.com/a/23905)
For benchmark [fuzzy-match](https://jsperf.com/fuzzy-match)

> Note: Search will be faster

lucacasonato:
I think we should disable search for queries < 3 chars. On slow Android devices searching is pretty painful otherwise.

rjoydip:
> I think we should disable search for queries < 3 chars. On slow Android devices searching is pretty painful otherwise.

Correct this will be good for mobile devices.

pages/x/index.tsx
sandeshdamkondwar: Does it make sense to sort every time query changes? Sorting algorithms are expensive compared to filter. And sorting is one time operation when entries are changed. That's the idea behind separating.

sandeshdamkondwar:
> I think we should disable search for queries < 3 chars. On slow Android devices searching is pretty painful otherwise.

Agree, will update the PR.

sandeshdamkondwar:
> > I think we should disable search for queries < 3 chars. On slow Android devices searching is pretty painful otherwise.
> 
> Agree, will update the PR.



> I think we should disable search for queries < 3 chars. On slow Android devices searching is pretty painful otherwise.

Gave a thought about this, not completely agree with this. People will search for terms like AB, API, URL, URI will end up in not finding what they want. Those terms generally goes inside description. Let's keep this way for now.

pages/x/index.tsx
sandeshdamkondwar: Both greedy a[^b]*b and possessive a[^b]*+b regexes are not working mentioned in [stackexchange link](https://codereview.stackexchange.com/a/23905).

`
const queryWords = query.split(" ").filter((word) => word.length);
queryWords.reduce((a, b) => {
        if (a === "") return b;
        return a + "[^" + b + "]*" + b;
}, "")
`

Tried here:
https://regex101.com/r/spw1H6/1
https://regex101.com/r/ecgzzN/1

@rjoydip  Do you have any solution for supporting backtracking?

lucacasonato:
> Gave a thought about this, not completely agree with this point. People will search for terms like AB, API, URL, URI will end up in not finding what they want. Those terms generally goes inside description. Let's keep this way for now.

Terms like `API` are 3 characters. I was talking about queries < 3 not <=3: `"API".length < 3 == false`.

About things like `AB` => these will match so many modules anyway because it is just part of a word, that it doesn't make any sense to search for this anyway.

pages/x/index.tsx
```diff
@@ -17,18 +17,22 @@ const ThirdPartyRegistryList = () => {
     setQuery(event.target.value);
   }
 
-  const list = useMemo(
+  // Sorting the list only if entries are changed
```
lucacasonato, line 20: Sorting shouldn't actually be needed because `database.json` is already in alphabetical order.

sandeshdamkondwar: @lucacasonato  Is there any automated job to update the list in database.json?

I don't see any. Devs might end up in writing anywhere in the list. 

Screenshot 2020-06-20 at 6 10 38 AM

lucacasonato: There should be a test for it.

sandeshdamkondwar: Nope, there is none. There is a prettier linting check, that's it.

lucacasonato: https://github.com/denoland/deno_website2/blob/master/util/database.test.ts#L50-L53

pages/x/index.tsx
rjoydip: > Does it make sense to sort every time query changes? Sorting algorithms are expensive compared to filter. And sorting is one time operation when entries are changed. That's the idea behind separating.

Sorting don't need https://github.com/denoland/deno_website2/pull/1171#discussion_r443072926
It has test case [database.test.ts#L50-L53](https://github.com/denoland/deno_website2/blob/master/util/database.test.ts#L50-L53)

pages/x/index.tsx
rjoydip: @sandeshdamkondwar I don't have any solution right now. I will try to look solution later.

ry:
I will wait for @lucacasonato's approval before landing

pages/x/index.tsx
sandeshdamkondwar: > https://github.com/denoland/deno_website2/blob/master/util/database.test.ts#L50-L53

Removing the sorting operation. Updated the PR.

pages/x/index.tsx
sandeshdamkondwar: Yes, will go that way. `a + ".*?" + b` is working fine. I'll revert back to that position. If you find any optimal solution, feel free to add in separate PR.

PR #1173, opened 2020-06-19
Make markdown images clickable

### LINKED ISSUE #1023: Allow images to be clickable to full size
The manual pages on the website are converted from markdown to html including the embedding of images.  These images are fixed width making screenshots hard to make out and no easy way to view them at full size.  For example, have a look at the screenshots on the [Debugger manual page](https://deno.land/manual/tools/debugger) and compare the screenshots with an example from this page [at original size](https://raw.githubusercontent.com/denoland/deno/master/docs/images/debugger3.jpg).

It would be good if the images could be clicked on to show them full size.

kidonng:
Resolves #1023
`target="_blank"` can be added but I will wait for requests.

### DIFF
--- a/components/Markdown.tsx
+++ b/components/Markdown.tsx
@@ -106,7 +106,11 @@ function ImageRenderer(props: { src: string; sourceURL: string }) {
     src = relativeToAbsolute(props.sourceURL, src);
   }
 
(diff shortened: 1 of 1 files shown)


kidonng:
I see, that's a problem with third party module's README.
Limit the scope to manual then?

PR #1181, opened 2020-06-19
Artwork page

### LINKED ISSUE #1162: Create artwork page
We're starting to collect a lot of Deno artwork - it would be good to have a place to collect and display them all.

Ideally each piece would have the artist with a link to their website, and the copyright license.

I've got a whole folder full of artwork from various artists - some of which hasn't been published online yet.

lucacasonato:
Closes #1162 

@ry Should this be linked to from anywhere other then the footer?


ry:
Thanks! Looks good - except can you make the json file a flat structure? Just to keep things dead simple for now.

For example I think this would be sufficient for now
```
    {
      "image": "/images/deno_logo.png",
      "alt": "a hand drawn deno logo",
      "artist": "https://github.com/ry",
      "license": "MIT"
    },
```

lucacasonato:
> can you make the json file a flat structure?

done

> I have a couple of artists who are not on Github. Maybe setting up profile pics is going to be hard for them

The fields, `github`, `twitter`, `web` and `profile_image` are optional. If they are not specified for an artist, it is gracefully handled by the UI. Also the `profile_image` can be a link to any image you want - it is not dependant on GitHub ID (I just happen to use the GitHub avatars for the artists currently on the page).

ry:
I've added some more artwork. One thing I came across is that sometimes the artwork itself has a website - like in the case of the 3d logo I want to link to https://github.com/MasterJames/high-res-deno-logo (cc @MasterJames) but there doesn't appear to be a way to do that.

Also maybe "title" instead of "alt" is more appropriate?

lucacasonato:
> I've added some more artwork. One thing I came across is that sometimes the artwork itself has a website - like in the case of the 3d logo I want to link to https://github.com/MasterJames/high-res-deno-logo (cc @MasterJames) but there doesn't appear to be a way to do that.

Good point. I'll add something for that in a minute.

> Also maybe "title" instead of "alt" is more appropriate?

`"alt"` is the accessibility description. It's not actually displayed in the UI - it's just for screen readers. I think it would make sense to also have a title though. I'll add it that.

ry:
What's the ordering? I think it would make most sense to order these chronologically - I can probably figure out approximate dates for most of these...

MasterJames:
In order to add the links to the 3D stuff I've contributed that pull request has to be merged first I guess.
If it is accepted and merged then I can add/link it to the website2 too if you prefer.
Let me know how it is best for me to help with that if needed.

MasterJames:
I was wondering where I should possibly repost the 3D version's Pull Request I made? https://github.com/denolib/high-res-deno-logo/pull/3

hashrock:
@MasterJames 
How about here?

https://github.com/denolib/3d-deno-logo

PR #1209, opened 2020-06-23
Artwork in rows

lucacasonato:
Display in rows rather than columns. Looks worse IMO, but the ordering is correct.

### DIFF
--- a/components/app.css
+++ b/components/app.css
@@ -61,31 +61,3 @@ ol.nested li:before {
 .highlight-line {
   background-color: #fffbdd;
 }
-.deno-artwork {
-  -webkit-column-count: 1;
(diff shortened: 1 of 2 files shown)


components/app.css
```diff
@@ -61,31 +61,3 @@ ol.nested li:before {
...
   background-color: #fffbdd;
 }
-.deno-artwork {
-  -webkit-column-count: 1;
-  -webkit-column-gap: 4rem;
-  -moz-column-count: 1;
-  -moz-column-gap: 4rem;
-  column-count: 1;
-  column-gap: 4rem;
-}
```
ry, line 71: Ah, this is what I was missing. Thanks!

PR #1213, opened 2020-06-23
Revert "fix: add padding to artworks"

lucacasonato:
Reverts denoland/deno_website2#1210

A unrelated change in this PR that changed`max-h-56` to `h-56` in makes wide image look squashed images on small screens: 
image

cc @ry @Qu4k


Qu4k:
My bad. I was experimenting a bit with the layout and forgot to revert.

Qu4k:
The change was made to align “An animated Deno logo” in the grid. Maybe we should look into aligning in better.

PR #1246, opened 2020-06-29
feat: enable check/no-check benchmarks

kitsonk:
With `--no-check` it makes sense to break out benchmarking information for TypeScript performance with a non-trivial workload.

### DIFF
--- a/pages/benchmarks.tsx
+++ b/pages/benchmarks.tsx
@@ -106,8 +106,8 @@ const Benchmarks = () => {
             <p className="mt-4">
               You are currently viewing data for{" "}
               {showAll ? "all" : "the most recent"} commits to the{" "}
-              <a href="https://github.com/denoland/deno">master</a> branch. You
-              can also view{" "}
+              <a href="https://github.com/denoland/deno">master</a>
+              branch. You can also view{" "}
               <Link
                 href="/benchmarks"
                 as={!showAll ? "/benchmarks?all" : "/benchmarks"}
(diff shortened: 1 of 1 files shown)


pages/benchmarks.tsx
```diff
@@ -173,6 +187,32 @@ const Benchmarks = () => {
...
+                <BenchmarkOrLoading
+                  data={data}
+                  columns={data?.execTime.filter(({ name }) => {
+                    console.log(name);
+                    return ["check", "no_check"].includes(name);
+                  })}
+                  yLabel="seconds"
+                />
+                <p className="mt-1">
+                  In both cases, a full type check is run for&nbsp;
```
lucacasonato, line 207: Why are all of the non breaking spaces needed?

kitsonk: Because with the formatting, you get things like "Withcheck a" when using things like `<em>`.

lucacasonato: Got it - you can replace all of the &nbsp; with `{" "}`. It serves the same purpose and prettier should intelligently add/remove these when needed. Non breaking spaces have the unintended side effect that they are non breaking :-)

kitsonk:
`--no-check` has been merged in Deno, and so meaningful benchmarks are now feeding the site.  (@ry it also shows that `--no-check` is saving about 35% of the full check time).

Benchmarks___Deno

pages/benchmarks.tsx
```diff
@@ -218,6 +232,31 @@ const Benchmarks = () => {
...
+                  columns={data?.execTime.filter(({ name }) => {
+                    console.log(name);
+                    return ["check", "no_check"].includes(name);
+                  })}
+                  yLabel="seconds"
+                />
+                <p className="mt-1">
+                  In both cases, <code>std/examples/chat/server_test.ts</code>{" "}
+                  is cached by Deno. The workload contains 20 unique TypeScript
+                  modules. With <em>check</em>a full TypeScript type check is
```
lucacasonato, line 254: Missing space between /em> and a.

PR #1258, opened 2020-07-01
Add DenoStars for deno

fakoua:


hacked:
This isn't a module, why should it be added to `database.json`?

fakoua:
you are right it is not a module.

PR #1270, opened 2020-07-03
Deprecate gh: and npm: redirects

lucacasonato:


components/Registry.tsx
```diff
@@ -180,6 +180,22 @@ const Registry = () => {
...
+                title="deno.land/x/gh:owner:repo redirect deprecation notice"
+                body={`The https://deno.land/x/gh:owner:repo style redirects are deprecated and will be removed on August 1st 2020. Instead of importing via deno.land you can import this module directly from GitHub${
+                  raw ? ` via the URL ${sourceURL}` : ""
+                }.`}
+              />
+            )}
+            {name.startsWith("npm:") && (
+              <WarningMessage
+                title="deno.land/x/npm:project redirect deprecation notice"
+                body={`The https://deno.land/x/npm:project style redirects are deprecated and will be removed on August 1st 2020. Instead of importing via deno.land you can import this module directly from unpkg.com via the URL ${
```
jamesmanning, line 194: @lucacasonato AFAICT you probably wanted to remove the " via the URL " after unpkg.com here since it'll get repeated?

lucacasonato: Thanks. Fixed in #1302

PR #1302, opened 2020-07-10
Deprecate npm

lucacasonato:
For some background: we are rewriting the deno.land registry service and it will initially not have NPM or arbitrary URL support. This might be re-added later. Only three entries are npm based so it does not have a high priority at the moment.

- `unexpected`, added by @alexjeffburke
- ~~`value-schema`, added by @shimataro~~ (#1303)
- `libauth`, added by @bitjson 

These entries will be removed on September 1st 2020.

### DIFF
--- a/components/Registry.tsx
+++ b/components/Registry.tsx
@@ -197,14 +197,15 @@ const Registry = () => {
                 }.`}
               />
             )}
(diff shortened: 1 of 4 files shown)


shimataro:
`value-schema` is fixed in #1303, but I hope NPM support will revive!

PR #1311, opened 2020-07-13
Add Apex Logs API client to the database

tj:


tj:
hmm I ran the formatter, we have to manually sort the database JSON?

tj:
Gotcha thanks man!

tj:
It would be nice to support `github.com/user/repo` instead of fighting over names in here, more like in Go:

```
import { Client } from 'https://github.com/apex/logs-js/client.ts'
``

crowlKats:
@tj you can import from github directly. with:
https://raw.githubusercontent.com/apex/logs-js/master/client.ts

PR #1332, opened 2020-07-19
ci: merge ci job added for automatically close PR where database and …

rjoydip:
This PR is made for add new CI Job (`merge`) to automatically close PR where database and artwork changes there.

**Condition**
- **Database changes:**
   - If there is only one file changed and of it is `database.json` then PR will auto-merge.
-  **Artwork changes:**
   - If there are two file changed one is `artwork.json` and image file for the artwork changes. then PR will auto-merge.

The merge job is dependence on `web` and `worker` job. Once both job gets success then only this job will work. Also it will merge only on `pull_request` event.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,3 +60,43 @@ jobs:
           environment: "staging"
           apiToken: ${{ secrets.cloudflareToken }}
           workingDirectory: worker
+  merge:
+    needs: [web, worker]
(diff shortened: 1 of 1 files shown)


ry:
Wwe've just now released an update that no longer uses database.json: https://deno.land/posts/registry2. As for the artwork, I think this complexity is too much for what it achieves. Thanks for the patch, but closing without merge.

PR #1370, opened 2020-07-28
Add GitHub link to error message

### LINKED ISSUE #946: Add a link to GitHub when rate limited
In the deno discord I saw a new user ask for help regarding std API because they were no longer able to view documentation. They didn't know that they could still view it if they look at the repository directly on GitHub instead of through the deno website.

I think we should include a link to the page on GitHub when a user encounters GitHub API rate limiting. I believe the same should apply for third party modules.

Below is a screenshot of the error.
Annotation 2020-05-30 172326

kidonng:
Add a handy link to GitHub if there is an error.

image

Fix #946

### DIFF
--- a/components/Registry.tsx
+++ b/components/Registry.tsx
@@ -327,7 +327,11 @@ function WarningMessage(props: { title: string; body: string }) {
   );
 }
 
-function ErrorMessage(props: { title: string; body: string; repositoryURL?: string }) {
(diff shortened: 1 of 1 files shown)


ry:
We've updated the deno.land/x service so this API issue should not appear anymore. 

See https://deno.land/posts/registry2

PR #1380, opened 2020-07-29
Add denopack module

jeroenptrs:
[https://github.com/denofn/denopack](https://github.com/denofn/denopack) Uses Rollup and Terser to bundle + minify code without using any Node related APIs/npm packages


ry:

jeroenptrs:
Thanks, @ry. Forgot this was even open still as I already tested out the new registry 😄

PR #1381, opened 2020-07-29
Blog page

lucacasonato:
A place for future release posts and announcements


components/Footer.tsx
```diff
@@ -53,9 +53,9 @@ const Footer = (props: { simple?: boolean }) => (
           </Link>
         </div>
         <div className="px-5 py-2">
-          <Link href="/v1">
+          <Link href="/blog">
             <a className="text-base leading-6 text-gray-500 hover:text-gray-900">
-              v1.0 Announcement
+              Blog
```
ry, line 58: News? Updates?

lucacasonato: News sounds good

pages/posts/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+              const date = new Date(post.publish_date);
+              const format = new Intl.DateTimeFormat(undefined, {
+                month: "long",
+                day: "numeric",
+                year: "numeric",
+              });
+              return (
+                <div key={post.id}>
+                  <p className="text-sm leading-5 text-gray-500">
+                    <time dateTime="2020-03-16">{format.format(date)}</time>
```
kidonng, line 61: Why is the `dateTime` attribute's value fixed? And also it should be `datetime`.

pages/posts/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+                  </p>
+                  <Link href="/posts/[article]" as={`/posts/${post.id}`}>
```
kidonng, line 63: Possibly should be `[post]` instead of `[article]`

PR #1421, opened 2020-08-04
Added subdirectory input validation

tajpouria:
Validate module subdirectory before getting `{"success":false,"info":"provided sub directory is not valid as it does not end with a /"}
` error on webhooks!
By the way with this implementation input likes '//' are valid. It just makes sure that either input does not provide or ends with '/' and I don't know if that's fine or not.

### DIFF
--- a/components/RegistryInstructions.tsx
+++ b/components/RegistryInstructions.tsx
@@ -230,7 +230,8 @@ function RegistryInstructions(props: { isOpen: boolean; close: () => void }) {
                             <p className="text-red-400 mb-2">
                               The provided subdirectory is not valid. It must
                               end with a <InlineCode>/</InlineCode>, but may not
-                              start with one. (e.g. <InlineCode>src/</InlineCode>)
+                              start with one. (e.g.{" "}
+                              <InlineCode>src/</InlineCode>)
                             </p>
                           ) : null}
                           <span className="text-gray-500">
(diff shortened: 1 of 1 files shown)


lucacasonato:
Thanks! I hadn't gotten around to this before release. I think the validation is actually that it may not start with a /, but must end in a slash. Could you update that in the validation regex?

tajpouria:
@lucacasonato Here you go!

lucacasonato:
@tajpouria Thanks, could you fix the lint error CI is complaining about? Then this is good to merge.

tajpouria:
@lucacasonato My bad!

components/RegistryInstructions.tsx
```diff
@@ -207,12 +214,24 @@ function RegistryInstructions(props: { isOpen: boolean; close: () => void }) {
...
+                            }`}
                             type="text"
                             placeholder="Subdirectory"
                             value={subdirectory}
                             onChange={(e) => setSubdirectory(e.target.value)}
                           />
+                          {isSubdirectoryValid === false ? (
+                            <p className="text-red-400 mb-2">
+                              Provided sub directory is not valid it should not
+                              start with a / and should end with a /
```
lucacasonato, line 232: Change to:

`The provided subdirectory is not valid. It must end with a <InlineCode>/</InlineCode>, but may not start with one. (e.g. <InlineCode>src/</InlineCode>)`

tajpouria: Changed

PR #1450, opened 2020-08-08
Move deno.land/x description under search

lucacasonato:


lucacasonato:
> That's better... but there's still something off about this 2 column layout. I suppose it will be worked out over time tho.
> 
> LGTM

Yeah I wanna rework it - I'm not gonna close your issue yet. This is just a temporary fix because I got fed up with having to scroll down to search.

PR #1454, opened 2020-08-10
feat: file search

lucacasonato:
You can now search an entire directory / module recursively for filenames. The shortcut to select the search bar is `t`. This is the same as on GitHub.

### DIFF
--- a/components/DirectoryListing.tsx
+++ b/components/DirectoryListing.tsx
@@ -1,0 +1,240 @@
+import React, { useMemo, useState, createRef, useEffect } from "react";
+import Link from "next/link";
+import { useRouter } from "next/router";
+import { isReadme, DirListing } from "../util/registry_utils";
+
+function DirectoryListing(props: {
+  dirListing: DirListing[];
+  name: string;
+  version: string | undefined;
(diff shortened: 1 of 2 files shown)


ry:
Cool! Why 't'? 

I would think '/' makes more sense. In docs.rs they use 'S' (always seemed odd to me)

lucacasonato:
> Why 't'?

It is the hotkey on GitHub so I thought people would be familiar with it. I can change it to `/` though.

SyrupThinker:
> It is the hotkey on GitHub

It's *a* hotkey (:
The slash on the search button says which one they expected to be used :D

Why not both?

ry:
Oh I didn't know about that feature on github. That makes sense. I'm ok with 't'.

lucacasonato:
> The slash on the search button says which one they expected to be used :D
>
> Why not both?

The `/` is for the full repo search in the top left. `t` is for what GitHub calls the `file finder`

SyrupThinker:
Weird, my buttons must be messed up, s opens the file finder for me :p
Ignore me then

PR #1457, opened 2020-08-11
fix: improved markdown rendering

### LINKED ISSUE #747: Html in markdown is ignored
Any HTML code in markdown is currently ignored:

https://github.com/denoland/deno_website2/blob/dbbe3783bafdcdbd3d384efebeab8fb24b3ac373/components/Markdown.tsx#L124

Many repos rely on it for aligning, sizing etc, therefore entire sections of their readmes/docs are lost on `deno.land/x/`.

Is this something you plan to support eventually?

lucacasonato:
Markdown rendering is now heavily improved. All side-effect free HTML will now be rendered. The styling also matches GitHub's markdown rendering now.

Fixes #747


components/Markdown.tsx
```diff
@@ -163,14 +58,106 @@ function Markdown(props: MarkdownProps) {
...
+    sanitizer: dompurify.sanitize,
+    highlight: (code, language) =>
+      renderToStaticMarkup(
+        <RawCodeBlock
+          code={code}
+          language={language as any}
+          disablePrefixes={true}
+          enableLineRef={false}
+        />
+      ),
```
lucacasonato, line 118: This isn't great. I still need to ponder on how to do this without including multiple highlighters / themes. I will revisit this later.

nayeemrmn:
Is it possible to get these changes on the home page as well? Inline code still looks not as clean for example. Seems like we have to used `marked()` there (for the main body of the page)

lucacasonato:
> Inline code still looks not as clean for example

I'll fix that

components/Markdown.tsx
ry: ok

PR #1467, opened 2020-08-13
Refactor /x

### LINKED ISSUE #1416: /x should display search and module listing above the fold
Currently there is a lot of text which isn't relevant to most users:
Screen Shot 2020-08-03 at 6 29 35 PM

The most interesting stuff should be visible. I propose moving the /x FAQ in part or completely to the manual or another page.

It would be nice to have the total module count above the fold too. Currently somewhat hard to find:

Screen Shot 2020-08-03 at 6 31 11 PM

littletof:
Closes #1416

- Adds module count into search placeholder
- Adds current page and total page count to mobile paginator

+ @lucacasonato suggested to move the `What is deno.land/x` section to the top, with a call to action to the other sections at the bottom of the page.
   - Not sure, if this is exactly what was suggested
   - I think the top would be a bit cluttered, if `Learn more` was a button, so I used a simple link.
   - It navigates to the bottom of the page with `#info`.

Theres is a problem, where if you navigate to the site with `https://deno.land/x#info`, then it navigates to the correct place, but then the modules load, so it scrolls to the top of the page. Don't know what I should do with that.

### DIFF
--- a/pages/x/index.tsx
+++ b/pages/x/index.tsx
@@ -73,7 +73,9 @@ const ThirdPartyRegistryList = () => {
               id="query"
               className="block w-full px-4 py-2 leading-normal bg-white border border-gray-200 rounded-lg outline-none shadow hover:shadow-sm focus:shadow-sm appearance-none focus:border-gray-300 hover:border-gray-300 mt-1"
               type="text"
-              placeholder="Search"
+              placeholder={
+                !resp ? "Search" : `Search through ${resp.totalCount} modules`
+              }
               value={query}
               onChange={handleSearchInput}
             />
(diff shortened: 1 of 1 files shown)


lucacasonato:
> Theres is a problem, where if you navigate to the site with https://deno.land/x#info, then it navigates to the correct place, but then the modules load, so it scrolls to the top of the page. Don't know what I should do with that.

@littletof I will look into this. Mind if I push to your branch to fix this?

littletof:
Thanks! 

> Mind if I push to your branch to fix this?

No, feel free to do anything with this PR 🙂

PR #1473, opened 2020-08-16
Format the header padding and max width

narwy:
From this:
image

To this:
image

### DIFF
--- a/components/Header.tsx
+++ b/components/Header.tsx
@@ -4,12 +4,22 @@ import React, { useState } from "react";
 import Link from "next/link";
 import Transition from "./Transition";
 
-const Header = ({ subtitle }: { subtitle?: string }) => {
+const Header = ({
+  subtitle,
+  widerContent,
+}: {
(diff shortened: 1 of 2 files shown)


narwy:
This also fixes the responsiveness issue as well @lucacasonato 

That said I think on the homepage with this change the `max-w-screen-sm` content sections could become `max-w-screen-md` and have a more reading area so they're less compact, opinions?

lucacasonato:
I think the header should only be wider on pages where the content is also wider (registry page). Everywhere else it should be as it is now.

As for generally changing max-w-screen-sm to max-w-screen-md: lets not right now. I need to think about this some more

narwy:
@lucacasonato You happy for a prop to be passed then, `widerContent` boolean? if true uses`max-w-screen-xl` if not defaults to `max-w-screen-lg` ?

narwy:
@lucacasonato over to you mate, added the `widerContent` if passed as true, does what it says on the box

PR #1484, opened 2020-08-19
Inline mobile paginator placeholder

littletof:
- Makes `/x` mobile paginator page placeholder inline
- Fixes `Previous`, `Next` buttons that changed size during loading

Before:
image

After:
image


lucacasonato (approved):
LGTM - thanks Szalay

PR #1485, opened 2020-08-19
Fix markdown code section horizontal scrollbar placement

littletof:
## Question
Looking at `.prism-code` sections on the main page, it doesnt have padding on the sides when scrolling:
image
Should the markdown code behave the same as `.prism-code` (and also Github)?

---

Horizontal scrollbar overlaps the last codeblock line. This PR fixes it.
There shouldn't be, and I couldn't see anything unrelated thing to change because of this.

Before:
image
image

After: 
image
image


lucacasonato (approved):
Thanks!

PR #1496, opened 2020-09-03
 Scroll element into view on hash change (#1494)

### LINKED ISSUE #1494: Line anchors are highlighted but not jumped to
Given an url such as https://deno.land/std@0.67.0/fmt/colors.ts#L230 the line is highlighted but the browser loads with the viewport at the top of the page.

Would be nice if it jumped to the anchor.

Galileo-dev:
The issue was caused by a delay between page load and code view loading. originally it tries to go to the anchor tag that hadn't been generated.

Fixes #1494


ry (approved):
Thanks

PR #1501, opened 2020-09-13
fix(status): replace "View On Github" link

qu4k:
Original link pointed to: https://github.com/denosaurs/debug/blob/0.1.1/ resolving to a 404
Edited link now points to: https://github.com/denosaurs/debug/tree/0.1.1/ resolving correctly


lucacasonato (approved):
LGTM - thanks Filippo

PR #1502, opened 2020-09-13
v1.4 release notes

lucacasonato:


public/posts/v1.4.md
```diff
@@ -0,0 +1,268 @@
...
+// Bad
+import { MyType } from "./mod.ts";
+export { MyType } from "./mod.ts";
+
+// Good
+import type { MyType } from "./mod.ts";
+export type { MyType } from "./mod.ts";
+```
+
+These flags also disallow the use of `const enum`.
```
nayeemrmn, line 160: This isn't true, is it?

lucacasonato: I am pretty sure they are meant to. If I understood @kitsonk correctly, `const enum` is a type directed emit. That actually may be another flag though...

public/posts/v1.4.md
```diff
@@ -0,0 +1,268 @@
...
+[VS Code extension for Deno](https://marketplace.visualstudio.com/items?itemName=denoland.vscode-deno)
+has had some major feature releases recently. Here is a quick summary:
+
+## Remote URL IntelliSense
+
+A great new feature of the extension is IntelliSense for deno.land imports. It
+gives you autocomplete suggestions for module names on deno.land/x, all of their
+versions, and their full directory listing. All of this is done without actually
+downloading the module source code, instead it is all powered by
+[the recent updates to deno.land/x](http://localhost:3000/posts/registry2).
```
littletof, line 253: I think this wasnt meant to link to `localhost` 🙂

public/posts/v1.4.md
nayeemrmn: They're not meant to. It used to be the behaviour that _ambient_ declarations threw under `isolatedModules` but even that is changed as of https://github.com/microsoft/TypeScript/pull/28465. Non-ambient `const enum`s never threw, they were always emitted like `enum` in that case. Note the following quote in that PR:

>  With non-ambient const enums, we know that an isolatedModules build will emit the
enum and produce a plain reference rather than inlining the constant, so
everything will still work.

This is the thing I complained was not documented anywhere, but there's no reason to believe it's a bug that is actually supposed to throw. I've seen it acknowledged many times in TS issues/PRs.

public/posts/v1.4.md
```diff
@@ -0,0 +1,268 @@
...
+this update. It is now faster and less buggy. Additionally the file size of
+dependencies is now displayed making it very easy to figure out what
+dependencies are adding a lot of code to your project.
+
+![a screenshot of running `deno info https://deno.land/x/brotli/mod.ts`, which prints the a module graph for the `https://deno.land/x/brotli/mod.ts` module](/posts/v1.4/info.png)
+
+## CSS styling in console.log
+
+Most modern browsers support styling `console.log` messages with CSS. In our
+ongoing effor to be as web compatible as possible, Deno now also supports CSS
```
bartlomieju, line 172: s/effor/effort

public/posts/v1.4.md
lucacasonato: :+1: I'll remove the line.

public/posts/v1.4.md
```diff
@@ -0,0 +1,268 @@
...
+styles to apply as an argument to `console.log`:
+
+```js
+console.log("%cStop!", "color:red;font-weight:bold");
+// This will print a bold red `Stop!` to the console.
+```
+
+Deno supports the CSS properties `color`, `background-color`, `font-weight`,
+`font-style`, `text-decoration-color`, and `text-decoration-line`. Support for
+these properties, and custom rgb, hex, and hsl colors depdend on if your
```
bartlomieju, line 185: drop `if`?

bartlomieju (approved):
LGTM

PR #1518, opened 2020-09-18
feat: endpoints for generic import IntelliSense

lucacasonato:
This will be used for the generic version of import IntelliSense in the VS Code extension.

This also adds the soon to be required `/.well-known/deno-import-intellisense.json` used for auto-discovery in the VS Code extension.


ry (approved):
LGTM

Issue #1527, opened 2020-09-26
Deno manual should link to code in version branch instead of master.


trivikr:
**Is your feature request related to a problem? Please describe.**
Deno manual links to code in master branch.

For example, the link to `lib.deno.unstable.d.ts` in [Standard Library](https://deno.land/manual@v1.4.1/standard_library) manual links to https://github.com/denoland/deno/blob/master/cli/dts/lib.deno.unstable.d.ts

Link: https://deno.land/manual@v1.4.1/standard_library
<details>
<summary>Screenshot</summary>

![Screenshot from 2020-09-26 14-43-39](https://user-images.githubusercontent.com/16024985/94350797-dfd4c880-0006-11eb-834f-b8c6e98d2030.png)

</details>

**Describe the solution you'd like**
Deno manual should link to code in version branch.
If the manual is being viewed for https://deno.land/manual@v1.4.1/standard_library
The `lib.deno.unstable.d.ts` should link to https://github.com/denoland/deno/blob/v1.4.1/cli/dts/lib.deno.unstable.d.ts

**Describe alternatives you've considered**
Selecting the branch tag from dropdown after realizing I'm not looking at the code for the specific Deno version

<details>
<summary>Screenshot</summary>

![Screenshot from 2020-09-26 14-48-31](https://user-images.githubusercontent.com/16024985/94350864-8b7e1880-0007-11eb-86f2-f3ffef7ecf72.png)

</details>

PR #1528, opened 2020-09-26
feat: update github code links in manual to version branch

trivikr:
Fixes: https://github.com/denoland/deno_website2/issues/1527

### DIFF
--- a/components/Manual.tsx
+++ b/components/Manual.tsx
@@ -495,11 +495,9 @@ function Manual() {
                     </svg>
                   </a>
                   <Markdown
-                    source={
-                      content
-                        .replace(/\$STD_VERSION/g, stdVersion)
-                        .replace(/deno\/blob\/master/g, `deno/blob/v${version}`)
-                    }
+                    source={content
+                      .replace(/\$STD_VERSION/g, stdVersion)
+                      .replace(/deno\/blob\/master/g, `deno/blob/v${version}`)}
                     displayURL={`https://deno.land/manual${
                       version ? `@${version}` : ""
                     }${path}`}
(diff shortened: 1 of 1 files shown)


components/Manual.tsx
```diff
@@ -495,7 +495,11 @@ function Manual() {
                     </svg>
                   </a>
                   <Markdown
-                    source={content.replace(/\$STD_VERSION/g, stdVersion)}
+                    source={
+                      content
+                        .replace(/\$STD_VERSION/g, stdVersion)
+                        .replace(/deno\/blob\/master/g, `deno/blob/v${version}`)
```
lucacasonato, line 501: I would rather we introduce a `$CLI_VERSION` string to replace, and use that in the manual. WDYT?

trivikr: I initially had thought about it after seeing `$STD_VERSION` in documentation - but this was an easier fix.

trivikr: Having said that, this fix will replace all links to master with `${version}` and may introduce bugs. Will replace `$CLI_VERSION` instead.

trivikr: Updated in [`b5cbb74` (#1528)](https://github.com/denoland/deno_website2/pull/1528/commits/b5cbb74e7e713d96993aa852554830cecbe478ca)

lucacasonato (approved):
LGTM, thanks. Now we just have to go through denoland/deno/docs and update the references there.

trivikr:
> LGTM, thanks. Now we just have to go through denoland/deno/docs and update the references there.

Here: https://github.com/denoland/deno/pull/7710

One issue with this is that code links will not be browse-sable from Markdown files.

lucacasonato:
> One issue with this is that code links will not be browse-sable from Markdown files.

:man_shrugging: - that is already the case with `$STD_VERSION`

PR #1530, opened 2020-09-28
Homepage: add import CDN 'esm.sh'

ije:


ry:
Thanks but this is not meant to be an exhaustive list of CDNs. I think the text is clear.

PR #1547, opened 2020-10-02
feat: copy markdown heading link to clipboard on click (#1524)

### LINKED ISSUE #1524: Copy anchor link to clipboard when clicking on anchor link in markdown heading

masterzenox:
The cleanest way I found to use an `onClick` event for heading links inside `dangerouslySetInnerHTML` is to listen for events on the whole `div` and then inside the callback check if the `event.target` has a class `octicon-link`.

To write to clipboard I'm using `Clipboard API` - [here's the caniuse](https://caniuse.com/mdn-api_clipboard_writetext). Should I implement a fallback using `document.execCommand("copy")`? [relevant caniuse](https://caniuse.com/mdn-api_document_execcommand_copy)

Closes #1524

### DIFF
--- a/components/Markdown.tsx
+++ b/components/Markdown.tsx
@@ -138,12 +138,21 @@ function Markdown(props: MarkdownProps) {
       <div
         dangerouslySetInnerHTML={{ __html: raw }}
         className={`markdown py-8 px-4 ${props.className ?? ""}`}
+        onClick={handleClick}
       />
     );
   } catch (err) {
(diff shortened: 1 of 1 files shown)


lucacasonato (approved):
Regarding compat: 2 year old Safari mobile wouldn't benefit much from this anyway (really a desktop feature), and IE11 is dead. So LGTM. 

Thanks for the contribution :-)

Issue #409, opened 2020-05-13
re-add dark mode


oplik0:
The new design is great, but unlike the older version it lacks dark mode.

I know that this is apparently worked on, but still - adding this issue to document this and perhaps inform others of the lack/status of this feature.

slikts:
Everything in my system is set to use dark mode, including web sites using [Dark Reader](https://darkreader.org/), but this doesn't work for deno.land, so now I have an actual headache from basically staring into a spotlight (it's late).

wojpawlik:
Dark Reader works for me on deno.land in Static and Filter(+) mode.

mumumilk:
I was thinking about submiting a PR for this issue, but before, i have a couple of questions:

- Do yall think it is a good idea to use tailwind's experimental dark mode feature? If so, should we do it based on `prefers-colors-scheme` only or build a toggle somewhere ?
- It is required to stick to default tailwind's gray palette?
- Should I submit a PR only when the whole docs are with dark mode or it can be a work-in-progress-style PR ?

lucacasonato:
> I was thinking about submiting a PR for this issue, but before, i have a couple of questions:
> 
> * Do yall think it is a good idea to use tailwind's experimental dark mode feature? If so, should we do it based on `prefers-colors-scheme` only or build a toggle somewhere ?
> * It is required to stick to default tailwind's gray palette?
> * Should I submit a PR only when the whole docs are with dark mode or it can be a work-in-progress-style PR ?

Hey @mumumilk, this issue is actually way more involved than it seems because we have user provided content on the website (e.g. https://deno.land/x/oak/README.md). Before starting on this we should discuss how we want user provided content to behave (what should it look like).

I think a good start would be to use the experimental dark mode feature in Tailwind with the `prefers-color-scheme` setting, and only style a few simple pages (`/artwork`, the homepage, `/benchmarks`). We can merge that and get some user feedback before we start styling markdown content.

sno2:
> 
> 
> > I was thinking about submiting a PR for this issue, but before, i have a couple of questions:
> > 
> > * Do yall think it is a good idea to use tailwind's experimental dark mode feature? If so, should we do it based on `prefers-colors-scheme` only or build a toggle somewhere ?
> > * It is required to stick to default tailwind's gray palette?
> > * Should I submit a PR only when the whole docs are with dark mode or it can be a work-in-progress-style PR ?
> 
[...]

PR #1548, opened 2020-10-03
Add dark mode

mumumilk:
Starts implementation of dark mode on docs as in #409 

routes affected:

/ (index)
   image

/artwork
image

/translations
image

/benchmarks
image

**Other routes such as `x` should not have dark mode.**

I also added an experimental palette called `grey`, so i wont modify light mode palette.


sno2 (commented):
The Cookie Banner could also do with a dark theme.  You probably didn't see it, though, as it stores whether it has been dismissed or not in the `LocalStorage`.  Otherwise, this looks great.

image

mumumilk:
@sno2 i left that white on purpose because on "light mode" it is black, contrasting with the rest of the website.

Should i fix conflicts for this pr? Its just aesthetic so i think its not a priority right now, so maybe we can close it. I'll be happy to help implementing this another time if so.

sno2:
> @sno2 i left that white on purpose because on "light mode" it is black, contrasting with the rest of the website.
> 
> Should i fix conflicts for this pr? Its just aesthetic so i think its not a priority right now, so maybe we can close it. I'll be happy to help implementing this another time if so.

Ahh, good explanation.  I understand.  Per the conflicts, you can fix them if you want, but but it looks like the higher ups aren't looking into this repo that much.

ry:
Hi @mumumilk - I'm going to close this because it's stale and out of date. Thanks for the effort - I apologize that we didn't give any encouraging reviews - it was a good start, just needed more work.
Cheers, Ryan

PR #1550, opened 2020-10-03
fix: set default cli version

tokiedokie:
When we don't provide specific version, $CLI_VERSION is undefined.
This cause some links to be invalid.

> https://deno.land/manual/runtime
> Additionally, a full list of the Web APIs which Deno implements is also available [in the repository](https://github.com/denoland/deno/blob/undefined/cli/rt/README.md).


lucacasonato (approved):
LGTM. Thanks

PR #1566, opened 2020-10-14
header: center Github icon with nav items

aykxt:
This should fix the incorrect position of the Github icon in the header


ry (approved):
Thank you - that has been bothering me

PR #1572, opened 2020-10-24
[Fix] [404] Footer links were not clickable

meszarosdezso:
Sorry for another PR, one of my friends noticed that on the 404 page the elements on the footer were not reacting to mouse/touch events.

This fixes it.

### DIFF
--- a/components/app.css
+++ b/components/app.css
@@ -115,11 +115,11 @@ ul.tree li:last-child:before {
 
 .NotFoundPage .animation {
   width: 100%;
-  height: 100%;
+  height: 400px;
   position: absolute;
   left: 0;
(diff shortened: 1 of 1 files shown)


lucacasonato:
One more thing I saw. It seems the page is always slightly to big to fit on the viewport (regardless of zoom). I always need to scroll down ~40px to view the entire screen. Do you know why this is?

meszarosdezso:
Thanks for noticing, I think I managed to fix the issue in #1578.
The problem was that I was using a hard coded height for the 'land', (so the gifs are in the correct place), but now that the footer is the land, I didn't encountered it's changing height. CSS flexbox to the rescue. (:

PR #1589, opened 2020-11-02
fix: date display

lucacasonato:


components/Registry.tsx
```diff
@@ -64,7 +64,7 @@ function Registry(): React.ReactElement {
   }
 
   function getFormattedDateTime(date: Date) {
-    const format = new Intl.DateTimeFormat(undefined, {
+    const format = new Intl.DateTimeFormat([], {
       hour: "numeric",
       minute: "numeric",
       second: "numeric",
```
ry, line 70: YYYY-MM-DD is non-ambiguous

lucacasonato: Maybe we just do `2 Nov 2020`? That seems the most obvious.

kt3k:
closing in favor of #1809

PR #1595, opened 2020-11-09
chore: disable std version prefix for >= v0.61.0

lucacasonato:


avihavai:
Hi! Disabling old std versions break libraries that have been released just a few months ago, including [https://deno.land/x/dejs@0.8.0](https://deno.land/x/dejs@0.8.0). What alternatives would you suggest to (1) moving away from the libraries that stop working due to the std version disabling and (2) avoiding deno.land and depending on github instead (e.g. [https://raw.githubusercontent.com/syumai/dejs/master/mod.ts](https://raw.githubusercontent.com/syumai/dejs/master/mod.ts))?

lucacasonato:
@avihavai We did not actually disable the std version. We only disabled the version prefixed `std` versions (so instead of importing `https://deno.land/std@v0.61.0` you must import `https://deno.land/std@0.61.0`). There has been a deprecation notice for std versions starting with the `v` prefix since July.

avihavai:
I know, sorry for not being explicit. It's a bit tricky as some library authors are not keeping track of the changes and still refer to v. Hopefully there will be a solution for this at some point. I completely agree that these should be deprecated and disabled at some point, although it's a good question whether a notice of a couple of months is sufficient.

lucacasonato:
@avihavai We tried to make it pretty clear by redirecting the frontend to the non `v` prefixed version, printing a warning in deno when you import a module using a `v` prefixed std version, and by disabling the std prefixes in batches instead of all at once. As Deno is only stable since may, (these v prefixes have been deprecated for more than half of the time since deno 1.0 was released), I think we gave enough warning. Sorry if this is impacting you

avihavai:
No problem, was just hoping for a meaningful way around this. Thanks for the great effort with deno :+1:

PR #1596, opened 2020-11-09
Add esm.sh CDN

ije:
pull request again, since [esm.sh](https://esm.sh) has better Deno compatibility:
- `X-Typescript-Types` header provided
- nodejs internal modules(fs,readling,etc) polyfill

```javascript
import postcss from 'https://esm.sh/postcss'
import autoprefixer from 'https://esm.sh/autoprefixer'

const css = (await postcss([ autoprefixer]).process(`
    backdrop-filter: blur(5px);
    user-select: none;
`).async()).content
```

### DIFF
--- a/pages/index.tsx
+++ b/pages/index.tsx
@@ -192,11 +192,11 @@ for await (const req of s) {
             ,{" "}
             <a href="https://jspm.io" className="link">
               jspm.io
-            </a>{" "}
+            </a>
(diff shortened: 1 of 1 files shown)


kt3k (approved):
@ije Thank you for your contribution, and sorry for the delayed review.

I agree that esm.sh is better for deno in many cases.

PR #1612, opened 2020-12-04
add Webi to installers list

coolaj86:
add Webi to installers list

### DIFF
--- a/pages/index.tsx
+++ b/pages/index.tsx
@@ -247,6 +247,17 @@ const InstallSection = () => {
       <CodeBlock language="bash" code={`brew install deno`} />
     </div>
   );
+  const webi = (
+    <div key="webi" className="my-4 text-gray-700">
+      <p className="mb-2">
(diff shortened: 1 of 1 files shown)


kt3k:
I'm afraid webi is too immature compared to other options, like homebrew, chololatey, scoop, etc, and therefore it would be confusing for new users if it's put in the front page...

ry:
Thanks for the effort but I agree with @kt3k

coolaj86:
Do you have some metrics of maturity that you'd like us to hit as targets?

The project is very mature at this point, and much less error prone than brew or apt, but ideally our users don't end up on GitHub - only when they have a problem (which is rare due to the maturity of the tools and the simplicity of the implementation), or something they they want is missing (hence we have almost as many contributors as stars).

I'm thinking of driving a campaign to get a thousand or stars on GitHub specifically so that it's more representative of the actual user base.

PR #1614, opened 2020-12-07
chore: deprecate all v prefixed std versions

lucacasonato:


ry:
Closing because stale.

PR #1617, opened 2020-12-10
add cf analytics

lucacasonato:
This adds Cloudflare Web Analytics. Cloudflare Web Analytics does not collect or use visitors’ personal data.


ry (approved):
LGTM

PR #1624, opened 2020-12-16
Fix header on news page

### LINKED ISSUE #1609: Posts page menu bug
Screenshot_2020-11-28-07-50-32

I think, solution somewhere in z-index.
UPD: latest Brave Browser on Android (which is based on Chrome)

mooxl:
This PR fixes #1609 by adding a z-index to the header component.

#### Before:
before

#### After:
after


lucacasonato (approved):
LGTM thanks

PR #1625, opened 2020-12-16
Hover transition effect on button

mooxl:
I noticed that every link on the index page is blue and gets a little lighter when hovering. Since I think the version button also falls into this category, it should have the same transition as the rest of the links to maintain the design line. 
With this PR the button gets the same hover transition as the rest of the links.


ry (approved):
Thanks

PR #1635, opened 2021-01-05
Add Spanish translation

JheysonDev:
I have translated a large part of the deno website which is served in https://deno-es.vercel.app


ry (approved):
Thanks

PR #1645, opened 2021-01-16
fix(benchmarks): update outdated links

Liamolucko:


ry (approved):
Thanks! Sorry for the slow review.

PR #1646, opened 2021-01-17
1.7 release post

lucacasonato:


public/posts/v1.7.md
```diff
@@ -0,0 +1,212 @@
+Today we are releasing Deno 1.7.0. This release contains many new features, some
+stabilizations, a some great improvements to existing APIs and tooling.
```
satyarohith, line 2:
```suggestion
stabilizations, and some great improvements to existing APIs and tooling.
```

wperron: to piggy back on this comment, the comma before "and" should also be removed

lucacasonato: >  the comma before "and" should also be removed

It is an oxford comma. Our previous blog posts generally have these. (although sometimes I _am_ a little liberal with my use of commas 😆 ) https://en.wikipedia.org/wiki/Serial_comma

public/posts/v1.7.md
```diff
@@ -124,7 +187,10 @@ running in a browser or not. This is bad practice. Use
 
 ## Improvements to coverage
 
-Bartek
+`deno test --coverage` had been improved to properly report partially covered
+lines as well as collect coverage for Deno subprocesses spawned from tests.
+We'll continue improving coverage capabilities including other report formats in
+the coming releses.
```
caspervonb, line 292:
```suggestion
the coming releases.
```

public/posts/v1.7.md
```diff
@@ -0,0 +1,423 @@
...
+{ "0": "A", "1": "B", "2": "C", A: 0, B: 1, C: 2 }
+0
+```
+
+Similarly for workers:
+
+```ts
+const promise = deferred();
+const tsWorker = new Worker(
+  "data:application/typescript;base64,aWYgKHNlbGYubmFtZSAhPT0gInRzV29ya2VyIikgewogIHRocm93IEVycm9yKGBJbnZhbGlkIHdvcmtlciBuYW1lOiAke3NlbGYubmFtZX0sIGV4cGVjdGVkIHRzV29ya2VyYCk7Cn0KCm9ubWVzc2FnZSA9IGZ1bmN0aW9uIChlKTogdm9pZCB7CiAgcG9zdE1lc3NhZ2UoZS5kYXRhKTsKICBjbG9zZSgpOwp9Owo=",
```
nayeemrmn, line 83: Perhaps this should be a demo using `btoa()`?
```suggestion
  `data:application/typescript;base64,${`
    if (self.name !== "tsWorker") {
      throw Error(\`Invalid worker name: ${self.name}, expected tsWorker\`);
    }

    onmessage = function (e): void {
      postMessage(e.data);
      close();
    };
  `}`,
```

public/posts/v1.7.md
```diff
@@ -0,0 +1,430 @@
...
+We had a lot of issue reported by users about random hangs occuring in numerous
+`Deno` APIs. After thorough investigation it was determined that all of them
+were caused by interaction with the `tokio` runtime. Due to API changes in
+`tokio` 1.0 we had to rearchitecture significant parts of `deno_core` to cater
+for those changes. In effect the `ResourceTable`, a structure that holds Rust
+allocated objects (eg. a file handle, TCP connection) was rewritten from
+scratch. We had added the ability to queue different "ops" acting on the
+resources, as well as graceful closing of those resources. That means that any
+pending reads would be canceled when closing a resource, while pending writes
+run until completion.
```
piscisaureus, line 418: Additionally - assuming https://github.com/denoland/deno/pull/9118 gets landed before the release:

* TLS streams can now send and receive concurrently.
* TLS streams can now be closed gracefully.
* UDP and UnixDatagram sockets can now send and receive packets concurrently.
* TCP, TLS, UnixStream `write` operations are serialized w.r.t other `write` operations. Similarly, `read` operations are serialized w.r.t. other `read` ops.
*  All I/O performed on a `File` resource is serialized.

piscisaureus: Didn't make the cut, unfortunately.

public/posts/v1.7.md
```diff
@@ -0,0 +1,465 @@
...
+[the TypeScript section of the manual](https://deno.land/manual@v1.7.0/typescript/runtime).
+
+## Markdown support in `deno fmt`
+
+`deno fmt` now supports formatting markdown files, including formatting of
+JavaScript and TypeScript codeblocks in these files.
+
+Additionally a new flag `--ext` was added to allowing specifying file extension
+when formatting code from stdin (`deno fmt -`). The available file extensions
+are: `js`, `jsx`, `ts`, `tsx`, `mjs`, `md`. Keep in mind that this flag has no
```
satyarohith, line 199:
```suggestion
are `js`, `jsx`, `ts`, `tsx`, and `md`. Keep in mind that this flag has no
```

The possible values for `--ext` doesn't include `mjs`.

public/posts/v1.7.md
```diff
@@ -0,0 +1,465 @@
...
+tsWorker.terminate();
+```
+
+## New unstable `Deno.resolveDns` API
+
+This release adds a new `Deno.resolveDns` API. It can be used to query DNS
+records from a DNS resolver. At the moment only DNS over UDP is supported (no
+DNS over HTTPS nor DNS over TLS). One can specify a custom nameserver (for
+example Cloudflare's 1.1.1.1 or Google's 8.8.8.8) to use, but by default we will
+use the system resolver (e.g. `/etc/resolv.conf` on Linux).
```
magurotuna, line 156:
```suggestion
This release adds a new `Deno.resolveDns` API. It can be used to query DNS
records from a DNS resolver. At the moment only DNS over UDP/TCP is supported (no
DNS over HTTPS nor DNS over TLS). One can specify a custom nameserver (for
example Cloudflare's 1.1.1.1 or Google's 8.8.8.8) to use, but by default we will
use the system resolver (e.g. `/etc/resolv.conf` on Linux).
```

Actually, lookup through TCP is also supported. (In fact, the integration test would pass with [this line](https://github.com/denoland/deno/blob/master/cli/tests/integration_tests.rs#L5553) commented out).
At first, I thought it would be good to have option for the users to specify which protocol to use. But it became impossible because I decided to use [this method of trust_dns_resolver](https://docs.rs/trust-dns-resolver/0.20.0/trust_dns_resolver/config/struct.NameServerConfigGroup.html#method.from_ips_clear) and it automatically creates both TCP and UDP connections internally.

magurotuna (changes requested):

public/posts/v1.7.md
```diff
@@ -0,0 +1,465 @@
...
+Below you can see an example of a simple hello world program being cross
+compiled for Linux from macOS, then it being run on Linux.
+
+<img src="/posts/v1.7/cross_compile.gif">
+
+Finally, `deno compile` can now create binaries that have built-in CA
+certificates, custom V8 flags, locked down Deno permissions, and pre-populated
+command line arguments. This should make `deno compile` useful to more people.
+
+Here is an example where we create a executable from the std/http/file_server
```
kt3k, line 64:
```suggestion
Here is an example where we create an executable from the std/http/file_server
```

public/posts/v1.7.md
```diff
@@ -0,0 +1,465 @@
...
+single improved function (`Deno.emit`). You can read all about how to use
+`Deno.emit` to bundle, transpile, and more in
+[the TypeScript section of the manual](https://deno.land/manual@v1.7.0/typescript/runtime).
+
+## Markdown support in `deno fmt`
+
+`deno fmt` now supports formatting markdown files, including formatting of
+JavaScript and TypeScript codeblocks in these files.
+
+Additionally a new flag `--ext` was added to allowing specifying file extension
```
kt3k, line 197:
```suggestion
Additionally a new flag `--ext` was added to allow specifying file extension
```

public/posts/v1.7.md
```diff
@@ -0,0 +1,465 @@
...
+
+// Send the body to the server.
+const res = await fetch("https://myfileserver.com/upload", {
+  method: "POST",
+  body,
+});
+```
+
+Currently the only runtimes supporting fetch upload streaming are Chromium and
+Deno. Both implementations have a limitation where you can not start recieving
```
kt3k, line 380:
```suggestion
Deno. Both implementations have a limitation where you can not start receiving
```

public/posts/v1.7.md
```diff
@@ -0,0 +1,465 @@
...
+For some more examples and usecases for fetch upload streaming take a look at
+[Jake Archibald's](https://twitter.com/jaffathecake) post on the matter:
+https://web.dev/fetch-upload-streaming/.
+
+## TLS session cache
+
+Some servers require that users reuse existing TLS sessions (for example FTP).
+Until now Deno has not had the ability to reuse TLS sessions, and instead would
+re-establish a new TLS session for each connection. In this release we have
+added a process global TLS session cache that will allows to reuse existing TLS
```
kt3k, line 394:
```suggestion
added a process global TLS session cache that will allow to reuse existing TLS
```

public/posts/v1.7.md
piscisaureus: recieving -> receiving

piscisaureus (approved):

PR #1651, opened 2021-01-21
Add news in header and menu

kt3k:
The news link is now only in the footer, and it feels so hidden. This PR adds the links to news in the header and the global menu.


lucacasonato (approved):
LGTM, thanks

PR #1652, opened 2021-01-29
Updated index.tsx

japrozs:
Improves the `deno_website2/pages/index.tsx`  file

Minor core changes

### DIFF
--- a/pages/index.tsx
+++ b/pages/index.tsx
@@ -30,7 +30,7 @@ for await (const req of s) {
       </Head>
       <CookieBanner />
       <div className="bg-white">
-        <div className="bg-gray-50 border-b border-gray-200">
+        <div className="bg-gray-50 border border-b border-gray-200">
           <Header />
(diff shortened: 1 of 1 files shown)


pages/index.tsx
```diff
@@ -30,7 +30,7 @@ for await (const req of s) {
       </Head>
       <CookieBanner />
       <div className="bg-white">
-        <div className="bg-gray-50 border-b border-gray-200">
+        <div className="bg-gray-50 border border-b border-gray-200">
```
kt3k, line 33: `border` is `border-width: 1px;`, `border-b` is `border-bottom-width: 1px;`. Specifying both doesn't make sense. (`border-b` is included in `border`)

kt3k:
I don't find enough reason for merging this. Closing without merge.

pages/index.tsx
japrozs: Sometimes tailwind doesn't render borders so its best to specify `border` seperately along with the classes

Issue #1592, opened 2020-11-06
Not found pages haven't response status code 404


eusonlito:
All not existing pages should return a `404` status code on response to allow to search engines and any request to `deno.land` to manage this errors as `404` errors and not as `200` success. For example, obsolete modules:

```
Download https://deno.land/x/postcss@8.1.2/lib/postcss.js
error: Uncaught (in promise) TypeError: An unsupported media type was attempted to be imported as a module.
  Specifier: https://deno.land/deno/postcss-deno-import/deno/mod.js
  MediaType: Unknown
```

`statusCode={404}` was removed on PR Custom 404 page 🦕 https://github.com/denoland/deno_website2/pull/1570

Duelsik:
#1654

Duelsik:
Created PR for that, fingers crossed 😄

crowlKats:
Since we switched to fresh, we provide 404 status with the caveat of https://github.com/denoland/dotland/issues/2258

PR #1654, opened 2021-01-30
Serve 404 page with 404 status code

Duelsik:
Aims to fix https://github.com/denoland/deno_website2/issues/1592

I am not next.js expert... so I advise to check if this is right place and lifecycle(?) to bind to.

### DIFF
--- a/pages/[...rest].tsx
+++ b/pages/[...rest].tsx
@@ -1,6 +1,7 @@
 /* Copyright 2020 the Deno authors. All rights reserved. MIT license. */
 
 import React, { useMemo } from "react";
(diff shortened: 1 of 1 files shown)


lucacasonato (changes requested):
Thanks for the contribution, but this is not something we will add, because it adds a serverless function execution to each page load.

If you programmatically request something from /x or /std, it is already served with a 404 response code.

lucacasonato:
Eg try `curl -i https://deno.land/x/does_not_exist@foo/bar`

Duelsik:
👍

PR #1657, opened 2021-02-02
fix: header has wider content on third party modules site

mooxl:
In the current version of the third party modules page, the navigation in the desktop view already has two-line links, because it is not wide enough. In this PR, the widerContent condition is set to true so that the links are no longer double-spaced.


kt3k (approved):
@mooxl This looks much better! Thanks for the improvement.

For reference:

Before
スクリーンショット 2021-02-04 18 40 58

After
スクリーンショット 2021-02-04 18 41 09

PR #1658, opened 2021-02-04
fix: wider header on benchmark page

mooxl:
Aims the same as the previous pull request (compare #1657), only that the header in the bechmark page is made wider, as this page has the same visual bug.


kt3k (approved):
@mooxl Thanks! LGTM

PR #1674, opened 2021-02-26
feat: Add husky

JheysonDev:


kitsonk:
Thank you, but we don't need precommit hooks and prefer not to have them.

Issue #1520, opened 2020-09-19
Force redirect from /x/std to /std


lucacasonato:

JheysonDev:
This issue has already been solved?

lucacasonato:
No

PR #1675, opened 2021-02-27
fix: Force redirect from /x/std to /std

JheysonDev:
this pull request solves the need for issue #1520 to force the redirect from /x/std to /std

### DIFF
--- a/pages/x/[...rest].tsx
+++ b/pages/x/[...rest].tsx
@@ -1,10 +1,15 @@
 /* Copyright 2020 the Deno authors. All rights reserved. MIT license. */
 
-import React from "react";
-
+import React, { useEffect } from "react";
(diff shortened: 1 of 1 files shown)


lucacasonato:
This should instead be done using redirects in the next config.

JheysonDev:
@lucacasonato I've corrected

JheysonDev:
@lucacasonato I've corrected

JheysonDev:
I have noticed that the current redirects in deno web do not work, because I would like to know if I could do how I send the pull request initially

ry:
@jheysonsaav Thanks for the PR but I'm going to close because this is stale and the CLA isn't signed. Please reopen a new one if you want to get this in still.

PR #1681, opened 2021-03-03
Fix typos in 1.8 post

avindra:
Spotted a few minor typos in the post, suggested fixes.

`deliminated` might be ok...  but it seems unusual. I am used to `delimited`.


lucacasonato (approved):

lucacasonato:
Thanks

PR #1689, opened 2021-03-03
Fix typos

kidonng:
- ~~Fix languages of some code blocks~~
- Add a missing colon
- Fix some typos in the Deno 1.8 blog post

### DIFF
--- a/pages/index.tsx
+++ b/pages/index.tsx
@@ -260,7 +260,7 @@ const InstallSection = () => {
     <div key="powershell" className="my-4 text-gray-700">
       <p className="mb-2">PowerShell (Windows):</p>
       <CodeBlock
-        language="powershell"
+        language="bash"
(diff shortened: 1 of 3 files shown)


pages/index.tsx
```diff
@@ -260,7 +260,7 @@ const InstallSection = () => {
     <div key="powershell" className="my-4 text-gray-700">
       <p className="mb-2">PowerShell (Windows):</p>
       <CodeBlock
-        language="bash"
+        language="powershell"
```
kt3k, line 263: This doesn't look improving the rendering of these code blocks, rather it seems introducing inconsistency of the `$` sign before each line. I'd prefer to keep these as is.

kt3k:
The colon fix and typo fixes look good to me.

pages/index.tsx
kidonng: I didn't notice this, makes sense!

ry (approved):
Thanks

PR #1690, opened 2021-03-03
Handle URL encoded parts of non-html requests

surgeboris:
VSCode's json language server encodes `@` as `%40` when fetching `$schema` for some json file (e.g. [one generated by `denon -i`](https://github.com/denosaurs/denon/blob/890a01ad8fae6ab294c270184c36a7ae4fcbbcc3/src/templates.ts#L15)).

Despite the fact that [browser handles `%40` just fine](https://deno.land/x/denon%402.4.7/schema.json), a fetch request gets status 404:
```
await fetch("https://deno.land/x/denon%402.4.7/schema.json", { "credentials": "omit", "mode": "cors" });
```

This PR attempts to fix that problem.

### DIFF
--- a/worker/src/registry.ts
+++ b/worker/src/registry.ts
@@ -7,7 +7,7 @@ export const S3_BUCKET =
 
 export async function handleRegistryRequest(url: URL): Promise<Response> {
   console.log("registry request", url.pathname);
-  const entry = parsePathname(url.pathname);
+  const entry = parsePathname(decodeURIComponent(url.pathname));
   if (!entry) {
     return new Response("This module entry is invalid: " + url.pathname, {
       status: 400,


lucacasonato:
Thanks, but we are not actually using the Cloudflare Worker at the moment due to an issue with them since yesterday, so we cant land this right now.

surgeboris:
@lucacasonato What are you using then? I can take a look at fixing it.

lucacasonato:
We are currently using a very very very hacking 4AM hot fix running fly.io. It is not public right now, so you won't be able to fix it. Once infra normalizes again I will address this.

surgeboris:
OK, thanks. Looking forward to see this bug fixed.

lucacasonato:
This is now fixed. For reference the code is now available in the `/proxy` directory in this repo.

PR #1697, opened 2021-03-04
fix build.settings again and use in memory cache instead of redis

lucacasonato:
The redis cache is very slow. We can use it as a secondary cache in the future, but for now just an in memory cache should do the trick.

### DIFF
--- a/proxy/deps.ts
+++ b/proxy/deps.ts
@@ -4,8 +4,5 @@ export type {
   Context,
   RouterContext,
 } from "https://deno.land/x/oak@v6.5.0/mod.ts";
-export {
-  Cache,
-  redisCache,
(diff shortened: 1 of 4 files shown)


proxy/deps.ts
```diff
@@ -4,8 +4,5 @@ export type {
   Context,
   RouterContext,
 } from "https://deno.land/x/oak@v6.5.0/mod.ts";
-export {
-  Cache,
-  redisCache,
-} from "https://deno.land/x/httpcache@0.1.0/redis.ts";
+export { Cache } from "https://deno.land/x/httpcache@0.1.0/redis.ts";
```
satyarohith, line 7:
```suggestion
export { Cache } from "https://deno.land/x/httpcache@0.1.0/mod.ts";
```

ry (approved):

PR #1700, opened 2021-03-08
chore: better error handling

lucacasonato:


proxy/main.ts
```diff
@@ -37,7 +46,16 @@ app.addEventListener("listen", (event) => {
 });
 
 app.addEventListener("error", (evt) => {
-  console.log("error", evt.error);
+  const ctx = evt.context;
+  if (ctx) {
+    const status = ctx.response.status;
+    console.log(
```
satyarohith, line 52: nit: maybe use `console.error` for errors?

satyarohith (approved):
Looks good.

proxy/main.ts
lucacasonato: It doesn't render any differently - I personally think `console.error` is pretty useless :upside_down_face:

satyarohith: I agree with the rendering part - I just thought it might be better to send errors to stderr for easy access.

lucacasonato: Ah I see - didn't know we print errors to stderr :upside_down_face:.

ry (approved):

PR #1707, opened 2021-03-11
fix: broken link in benchmarks

trainto:
Due to https://github.com/denoland/deno/pull/9493,
Link to worker round robin test from benchmarks page has been broken.
This fixes that.


kt3k (approved):
Thanks! LGTM

PR #1708, opened 2021-03-11
Display relative time for modules

kidonng:
Resolve #1589

I looked at how other websites display the dates:

npm:

image

Yarn:

image

They are both using relative dates, which in my opinion is more friendly since you instantly know how up to date the package is. As for the tooltip (`title`) I choose `toLocaleString` just like Yarn because it follows user's system settings and should be the format they expect.

image


ry (commented):
LGTM - thanks @kidonng - sorry for the slow review!

ry (approved):

PR #1709, opened 2021-03-11
Correctly handle module version

### LINKED ISSUE #1718: #0 tag results in module not loading
Opening this url: https://deno.land/x/dxi

infinitely appends `#0`s, resulting in an url like this: `https://deno.land/x/dxi@#0#0#0#0#0#0#0#0#0#0...` and fails to load the module.

I noticed, that the module is an empty repo and also has a `#0` tag, maybe that causes the problem:

image

kidonng:
[xstate](https://github.com/davidkpiano/xstate) - this module is painful because its version has @ (at) and / (slash), which breaks literally everything:

- deno.land website
- deno.land proxy
- cdn.deno.land
- doc.deno.land

Yet this module is the 5th one on deno.land/x, which is why I spotted this issue.

- This PR fixes the website
- I will submit a PR to doc.deno.land
- I don't know how cdn.deno.land exactly works, but `getVersionDeps` fails on both escaped URL (https://cdn.deno.land/xstate/versions/xstate%404.16.2/meta/deps_v2.json) and unescaped (https://cdn.deno.land/xstate/versions/xstate@4.16.2/meta/deps_v2.json)
- Unfortunately for the proxy, I found oak decodes the URL before passing it to path-to-regexp, so the router can't match the correct version even if we escaped it. I will submit a PR to them as well.

P.S. dealing with all these paths is a nightmare, I think a helper function for creating these URL is needed or there will be more issues in the future.

Close #1718

### DIFF
--- a/components/DirectoryListing.tsx
+++ b/components/DirectoryListing.tsx
@@ -3,7 +3,7 @@
 import React, { useMemo, useState, createRef, useEffect } from "react";
 import Link from "next/link";
 import { useRouter } from "next/router";
-import { isReadme, DirListing } from "../util/registry_utils";
+import { isReadme, DirListing, getBasePath } from "../util/registry_utils";
 
 function DirectoryListing(props: {
   dirListing: DirListing[];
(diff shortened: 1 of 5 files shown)


kt3k:
@kidonng Nice refactoring and fixes!

But can you add test cases of `getRepositoryURL` (with 3rd arg), `parseNameVersion`, `getBasePath` in `util/registry_utils.test.ts`?

> Unfortunately for the proxy, I found oak decodes the URL before passing it to path-to-regexp, so the router can't match the correct version even if we escaped it. I will submit a PR to them as well.

Can you describe a bit more about what happen for which path?

wperron:
FYI, there's this related issue on the registry https://github.com/denoland/deno_registry2/issues/216

kt3k (approved):
LGTM

kidonng:
Sorry for not responding for a long time!

> @kidonng Nice refactoring and fixes!
> 
> But can you add test cases of `getRepositoryURL` (with 3rd arg), `parseNameVersion`, `getBasePath` in `util/registry_utils.test.ts`?

I will try to find time to add these in the future 😀

> > Unfortunately for the proxy, I found oak decodes the URL before passing it to path-to-regexp, so the router can't match the correct version even if we escaped it. I will submit a PR to them as well.
> 
> Can you describe a bit more about what happen for which path?

The issue is that `/` is always treated as a path separator, and can't be matched. Will link create a issue there and link it to here, again, if I find time.

PR #1713, opened 2021-03-22
style: add border radius to .prism-code

trainto:
I added a bit of radius to .prism-code, and IMO it looks better especially on mobile.
But I don't mind if this PR will not be accepted. I'm just suggesting. 😄


kt3k (approved):
The code blocks in the doc site already have small border-radius value. So this change makes sense to me.

LGTM

PR #1714, opened 2021-03-23
Update README.md

argos44:
I've seen the linux games.... so sad... getting killed on an orbitaal base above a burnt out planet.... COBOL was the language... they did not camouflage themselves COMRADES, HEYY!

but... then you might just get..... by friendly fire


argos44:
think about it... they did leave a time capsule... there was a meaning in it... it was a meaningful end......

PR #1719, opened 2021-03-29
fix: fix for empty module

### LINKED ISSUE #1718: #0 tag results in module not loading

kt3k:
This PR fixes the case when the repository doesn't include any file (which is the case of `dxi@#0`). Now `dxi#0` looks like the below in the registry.

https://deno-website2-b1tv0i1ri-denoland.vercel.app/x/dxi

スクリーンショット 2021-03-29 12 22 58


kidonng:
Looks like this is the same issue I'm trying to solve in #1709 (except I didn't fix the empty list issue which is another bug).

kt3k:
@kidonng Thanks for pointing. Ok. Let's work on that one first.

ry (approved):

PR #1723, opened 2021-03-30
Add missing redirect changes

kidonng:
This link is left out in #1720.


kt3k (approved):
@kidonng Good catch. Thanks!

PR #1735, opened 2021-04-15
remove source link of http_bench_bin_ops

justjavac:
https://github.com/denoland/deno/pull/10145


satyarohith:
A re-run should fix the ci.

(cc @kt3k)

kt3k (approved):
@justjavac Thanks for the fix!

PR #1736, opened 2021-04-15
favicon for safari

Gumichocopengin8:
- Add favicon for safari because favicon on Safari is not visible.

### DIFF
--- a/pages/_app.tsx
+++ b/pages/_app.tsx
@@ -14,8 +14,8 @@ export default class DenoWebsiteApp extends App {
     return (
       <div className="h-screen">
         <Head>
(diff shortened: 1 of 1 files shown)


Gumichocopengin8:
Chrome some take .ico favicon instead of .svg. In order to fix this, added size attribute. 
Bug details https://bugs.chromium.org/p/chromium/issues/detail?id=1162276

Gumichocopengin8:
@kt3k Could you review this PR too when you have time? 
Thanks.

kt3k (approved):
@Gumichocopengin8 LGTM. Thanks!

PR #1764, opened 2021-05-28
chore: better error handling

lucacasonato:
Nicer logging, and don't crash the isolate on errors.


wperron (approved):
The color is a nice touch :+1:

PR #1772, opened 2021-06-10
fix: retry proxy errors

lucacasonato:


worker/handler.ts
```diff
@@ -88,12 +88,25 @@ export function extractAltLineNumberReference(
...
   };
   const urlR = remoteUrl + url.pathname;
   const modifiedRequest = new Request(urlR, init);
-  return fetch(modifiedRequest);
+  let lastErr;
+  for (let i = 0; i < 3; i++) {
+    try {
+      return await fetch(modifiedRequest);
+    } catch (err) {
+      console.warn("retrying on proxy error", err);
```
wperron, line 107: maybe a little delay here?

lucacasonato: Delay should not be needed - it would only slow down the user connection. The failure case we are trying to fix by this is not timing based, so an extra deploy would not help.

piscisaureus: I think we should check at this point whether the error is what we expect.
But I guess it needs to land so we can run it. So add a todo?

piscisaureus (approved):
👠

Issue #1768, opened 2021-06-03
Add support for org-mode readmes


GuiltyDolphin:
Currently there appears to only be support for markdown readmes. If you try to use any other readme (e.g., Org), then no readme file will be detected, and documentation will not be displayed.

I propose that the website support Org readmes. I know this is much less common than markdown, but I tend to write my readmes in Org and was caught out when I found that the readmes weren't being detected.

Relevant files:

* https://github.com/denoland/deno_website2/blob/main/components/Markdown.tsx - this deals with most of the conversion to HTMl, and uses =marked= (https://marked.js.org/) to do the parsing.\
* https://github.com/denoland/deno_website2/blob/0b3b183ba44fdf7555edafe958b93f84fd7671c3/util/registry_utils.ts - this contains functions such as `findRootReadme` and `isReadme` that will need updating

I think the main thing to decide upon is which org-mode parser to use. For example:

* https://github.com/daitangio/org-mode-parser
* https://github.com/mooz/org-js
* https://github.com/glmxndr/orgjs (probably not this one)

etc.

I'm happy to have a shot at implementing this, but considering there are some choices to be had re: which parser to use, I would greatly appreciate input on which parser would be good to go with. Has anyone used any of these (or other) parsers in your own projects? How have you found them in terms of speed and ease of use? Any rules and regs on what sorts of packages Deno can import?

kt3k:
`mooz/org-js` looks better to me. Bundle size is smaller (7kb) and it has no external dependencies. `org-mode-parse` depends on node.js native modules like `fs` and `util`. I think this isn't suitable for frontend usage.

GuiltyDolphin:
@kt3k Yep, `mooz/org-js` looks pretty good from a dependency standpoint. I do worry though that it hasn't been updated since 2015, so it might not keep up-to-date with new features in org.

We could potentially fork that and try and maintain it with new features of Org, but I don't know how much work this would be. I suspect it probably supports most features that would be wanted, but it would be ideal if there was a more up-to-date package.

kt3k:
This one looks more maintained (and linked from the official website), https://github.com/orgapp/orgajs, the usage looks more complex than mooz/org-js though.

GuiltyDolphin:
That looks pretty good. I've got quite a bit of experience working with ASTs, so it shouldn't be an issue in terms of complexity. I think the main difficulty will be in rendering, but I'm hoping I can (for the most part) make use of the markdown styling for that.

[...]

PR #1779, opened 2021-06-23
Add support for rendering Org readmes

GuiltyDolphin:
This is regarding issue #1768.

Sample project for testing capabilities is `orgreadmetest` (https://deno.land/x/orgreadmetest).


GuiltyDolphin:
Getting there!

https://deno-website2-biejkizm6-denoland.vercel.app/x/functionality@v0.2.3:

image

Still some WIP though, e.g., code blocks:

https://deno-website2-biejkizm6-denoland.vercel.app/x/dentest@v0.1.0:

image

And some wrapping needs fixing:

https://deno-website2-biejkizm6-denoland.vercel.app/x/dspv@v0.1.0:

image

GuiltyDolphin:
@kt3k It looks like the issue with list content rendering weirdly is to do with the parser rather than the rendering itself. There is an open issue (https://github.com/orgapp/orgajs/issues/91), so I wonder whether it's better to wait for a fix in `orgajs` and then update the code here if needed, or if we should just build a workaround in (not super trivial) here.

Either way, I think we can probably integrate this before a fix as it can always be added later.

Still got some bits to work on, but I'll let you know when I think the code is ready for review. Thanks!

GuiltyDolphin:
Inline footnotes are currently not supported by `orga`. I've made a pull request to add this (https://github.com/orgapp/orgajs/pull/96).

GuiltyDolphin:
I think the only thing left to do (that doesn't need changes in orga) is checkboxes - not sure how we want to style these, but we can always just render as text if need be.

GuiltyDolphin:
@kt3k Ready for review!

Latest version can be seen working at https://deno-website2-lq25ldh19-denoland.vercel.app/x/orgreadmetest@v0.1.3

lucacasonato:
Cool that this works, but I have some questions/concerns:

- I ran lighthouse on this and compared to `main` we load about 50KB more JS for a /x pageload now (including ones that do not render `org`, which is most). That is an increase of around 40-50%. Orga alone is [28.9kb minified](https://bundlephobia.com/package/orga@2.4.8).
- The addition of `date-fns`: why does a markup renderer require a date formatter / parser?
- I am not so thrilled that we now have both `marked` and `reorg` as dependencies. They both cover very similar feature sets.

kt3k:
> I ran lighthouse on this and compared to main we load about 50KB more JS for a /x pageload now (including ones that do not render org, which is most). That is an increase of around 40-50%. Orga alone is 28.9kb minified.

`next build` command shows this increases 120kb -> 141kb first load JS for `/x/[...rest]` page. I don't think this affects that much. (Chrome devtool network tab with JS filtering also shows 169kb -> 195kb load, 26kb/15% increase)

But it's true that most pages don't use .org format. Maybe we can use `next/dynamic` for loading `Org` component lazily. What do you think?

lucacasonato:
@kt3k DevTools showed an extra 50KB bundle for me 🤷‍♂️

GuiltyDolphin:
I'll have a look into minimizing the dependencies.

GuiltyDolphin:
@lucacasonato 


>     * I ran lighthouse on this and compared to `main` we load about 50KB more JS for a /x pageload now (including ones that do not render `org`, which is most). That is an increase of around 40-50%. Orga alone is [28.9kb minified](https://bundlephobia.com/package/orga@2.4.8).

I've switched to lazy loading of the `Org` module, which seems to save space on general page loads (`122kb` first load vs `120kb`, now). I only actually need to use `orga` (directly) for the types - `reorg-parse` is used for the actual parsing. I'm looking into whether I can just require the types files from `orga`.

>     * The addition of `date-fns`: why does a markup renderer require a date formatter / parser?

See https://github.com/orgapp/orgajs/issues/83#issuecomment-826401251. But again, if I can lose the `orga` direct dependency, then maybe we won't need this.

>     * I am not so thrilled that we now have both `marked` and `reorg` as dependencies. They both cover very similar feature sets.

Well `orga` uses `unified`, so we could potentially use a markdown parser from there - might lead to less extra dependencies.

GuiltyDolphin:
@lucacasonato @kt3k I think that's as small as I can gct the deps. Set them to load only when needed though, so we only incur the ~50kb when using a project with an org readme.

package.json
```diff
@@ -32,12 +33,10 @@
     "react-dom": "npm:@preact/compat",
     "react-ssr-prepass": "npm:preact-ssr-prepass",
     "react-transition-group": "^4.4.1",
-    "reorg-parse": "^2.3.16",
     "swr": "^0.4.2",
     "tailwindcss": "^1.9.6",
     "tailwindcss-hyphens": "^0.1.0",
     "timeago-react": "^3.0.2",
-    "unified": "^9.2.1",
```
kt3k: 👍

kt3k (approved):
LGTM

kt3k:
@GuiltyDolphin Thank you for your contribution! Nice work!

PR #1780, opened 2021-06-26
Eyecandy: Longer transitions

Milo123459:
On the index, I have decided to make the transitions longer. It gives the website a much more relaxed feel. 

Hopefully, this is premium eye candy!

PR #1781, opened 2021-06-28
Revert "Eyecandy: Longer transitions"

lucacasonato:
Reverts denoland/deno_website2#1780

@Milo123459 Thanks for the PR, but I think this makes the whole site feel slower. I am going to revert it.


wperron:
> I think this makes the whole site feel slower.

I think this is a bit harsh, the patch literally just changes the duration of the animation on-hover for a couple of clickable items

lucacasonato:
That's the power of UI! Even minuscule changes can make or break it. I think it makes the experience of navigating around the site worse 🤷‍♂️

lucacasonato:
@Milo123459 No worries. Nothing against you personally!

kitsonk (approved):
LGTM

PR #1800, opened 2021-07-20
Update Installations part adding winget

AngelNext:
A very small change that adds in the webpage winget as an install method!

### DIFF
--- a/pages/index.tsx
+++ b/pages/index.tsx
@@ -269,6 +269,12 @@ const InstallSection = () => {
       />
     </div>
   );
+  const winget = (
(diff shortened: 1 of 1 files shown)


lucacasonato:
Deno was removed from winget because it uses an unofficial source.

https://github.com/microsoft/winget-pkgs/commit/090872823fb3110be70b2d612e6dfc210ec0add6#diff-5baa728b81954154b1cda21b41179dc9b5d7003e23c17b7220122b9c33b597ac

@piscisaureus do we care about actually getting the binary on WinGet?

Issue #1766, opened 2021-05-31
Make title of manual pages more descriptive


takenspc:
Hi, I am learning Deno little by little. Though manual pages provide various knowledge, I found that it is hard to identify the right page in browser tabs, histories, bookmarks and so on as every page has same title ("Manual | Deno").

![](https://user-images.githubusercontent.com/270905/120158785-76f5be00-c22f-11eb-9a25-abb2fb4b84d0.png)
 
In other languages such as Rust ([The Book](https://doc.rust-lang.org/book/)) and Node.js ([Guides](https://nodejs.org/en/docs/guides/)), pages have (semi-) unique titles so that I can quickly identify the right page.

Thus, I would like to make manual pages more descriptive. Based on Rust Book, I think following format is suitable for manual pages.

```html
<title>(Page name shown in sidebar) | Manual | Deno (Version)</title>
```

For example, https://deno.land/manual/typescript/migration will have following title:

```html
<title>Migrating to/from JavaScript | Manual | Deno v1.10.2</title>
```

Notes:

*   Though "Migrating to/from JavaScript" differs first heading of the page content ("Migrating to and from JavaScript") little bit, it provides good enough context for readers. I' not sure why two differ.
*   While "Migrating to/from JavaScript" is a subtopic of "Using TypeScript", "Using TypeScript" is omitted from the title to make title less verbose.

Proof of concept code is following:

*    https://github.com/takenspc/deno_website2/compare/manual-title

If the format is OK, I'll submit PR.

kt3k:
Thank you for looking into this. The change almost looks good to me.

I think the middle `Manual` is redundant for the case of `Migrating to/from JavaScript | Manual | Deno v1.10.2`. So maybe it's better to make it `Migrating to/from JavaScript | Deno v1.10.2` in my view.

PR #1802, opened 2021-07-21
fix: Make title of manual pages more descriptive

hirasawayuki:
resolve https://github.com/denoland/deno_website2/issues/1766

In this PR, all Manual pages were labeled `Manual | Deno`, so the titles of the Manual pages were made more descriptive.

When the state of the sourceURL is changed, the title of the page is also changed.

The attached image is the browser history.
before | after
---- | ----
before | after

### DIFF
--- a/util/manual_utils.test.ts
+++ b/util/manual_utils.test.ts
@@ -1,6 +1,6 @@
 /* Copyright 2020 the Deno authors. All rights reserved. MIT license. */
 
-import { getFileURL, getTableOfContents } from "./manual_utils";
+import { getFileURL, getTableOfContents, getPageTitle } from "./manual_utils";
 import "isomorphic-unfetch";
 
 /* eslint-env jest */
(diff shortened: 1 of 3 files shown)


bartlomieju:
A nitpick from me: maybe `{title} | Manual | Deno` would be better to clearly signal that this is a manual page?

hirasawayuki:
I'll fix it.

hirasawayuki:
@bartlomieju 
Thank you for your suggestion. I have fixed it!

bartlomieju:
Thanks! I'll wait for Luca's review before merging.

lucacasonato (commented):
Can we instead use the page title from the table of contents?

hirasawayuki:
@lucacasonato 
Thanks for the review!
I fixed it to get the page title from the table of contents.

util/manual_utils.test.ts
```diff
@@ -30,3 +34,19 @@ test("get introduction file new repo", async () => {
     "https://deno.land/x/manual@v1.12.1/introduction.md"
   );
 });
+
+test("get page title", async () => {
```
kt3k, line 38: :+1:

kt3k (approved):
Nice work! LGTM. Thank you for the contribution @hirasawayuki

Issue #231, opened 2020-02-07
[feat] hide hidden files in directory listing


brandonkal:
These files can still be served from the proxy and via direct link, but it would make sense to exclude these files from the listing, maybe adding a button at the bottom of the list:
"Show 3 hidden items"

divy-work:
I was looking forward to contributing or is there a PR already?

lucacasonato:
@divy-work This has not been taken by anyone yet, so feel free to work on it.

crowlKats:
This was implemented in the linked PR.

PR #1807, opened 2021-07-23
feat: hide hidden files in directory listing

hirasawayuki:
feat: https://github.com/denoland/deno_website2/issues/231

In this PR, I've adds feature to show/hide hidden files on the `Standard Library` and `Third Party Modules ` pages.

Click on `show hidden items` to toggle between showing and hiding.

demo
スクリーンショット 2021-07-23 12 03 42

### DIFF
--- a/components/DirectoryListing.tsx
+++ b/components/DirectoryListing.tsx
@@ -3,7 +3,12 @@
 import React, { useMemo, useState, createRef, useEffect } from "react";
 import Link from "next/link";
 import { useRouter } from "next/router";
-import { isReadme, DirListing, getBasePath } from "../util/registry_utils";
+import {
+  isReadme,
+  DirListing,
+  Entry,
+  getBasePath,
+} from "../util/registry_utils";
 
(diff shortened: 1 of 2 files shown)


kt3k:
The preview deployment looks working great! nice work @hirasawayuki 

But I'm not sure this is what users expect. For example, github doesn't hide these items by default. I'm afraid it could be confusing to the users.

I'd like to hear other collaborators' opinions as well.

satyarohith:
Nice work, @hirasawayuki!

.@kt3k, I'm +1 for landing this feature. It makes sense for GitHub not to hide these items because it represents a git repository. On the other hand, we represent a deno module, for which the dotfiles don't seem to provide much value to a user using the deno module.

kt3k:
Now I agree with Satya. Thank you for the input.

BTW I've also seen a similar request to this in twitter https://twitter.com/KawarimiDoll/status/1419122343327068161

kt3k (approved):

PR #1808, opened 2021-07-24
feat: Add localisation to "Version Info" field on module page

AnInternetTroll:


AnInternetTroll:
I used npm instead of yarn sorry

PR #1809, opened 2021-07-24
feat: Add localisation to "Version Info" field on module page

AnInternetTroll:


AnInternetTroll:
My PR seems related to #1589, which I hadn't seen before opening mine.

kt3k:
hmm seems like this doesn't work for users with `navigator.language === "en"` ref:https://github.com/denoland/deno_website2/pull/1589#discussion_r516239510

AnInternetTroll:
> hmm seems like this doesn't work for users with `navigator.language === "en"` ref:[#1589 (comment)](https://github.com/denoland/deno_website2/pull/1589#discussion_r516239510)

I'm not sure I understand what you mean, it works fine for me: 
image

AnInternetTroll:
One thing that could be done would be to change `month: "numeric",` to `month: "short",`. Only issue with this is that the month name will appear in the user's language, so if I were to be German it would tell me `26. Juli 2021` which might be confusing given the rest of the site is in American English.

kt3k:
> One thing that could be done would be to change month: "numeric", to month: "short",. Only issue with this is that the month name will appear in the user's language, so if I were to be German it would tell me 26. Juli 2021 which might be confusing given the rest of the site is in American English.

`month: "numeric", to month: "short` sounds good to me, and because the web site already hard code English everywhere how about hard code locale "en" in the constructor?

kt3k (approved):
Thanks! LGTM

PR #1817, opened 2021-08-09
fix: tell algolia the version of a given doc page

lucacasonato:
I am hoping this fixes the duplicate results in search.


lucacasonato:
Algolia only scrapes the latest version now.

PR #1826, opened 2021-08-16
deno.land Version Aliasing

### LINKED ISSUE #606: Allow semver ranges and identifiers in urls
The website should be able to resolve semver symbols. Being able to use semver to dedupe dependencies is an important step towards facilitating the development of medium to large web apps using Deno. I believe some other github redirectors already support this as well.  While we could use those CDNs instead, the deno.land registry is currently the only place that exclusively lists deno modules. It may be possible to build further deduping tools once basic semver support is in place.

tani:
This pull request closes #606 .

lib.deno.dev provides a version aliasing for deno libraries.

```typescript
// node-semver style
import abc from "https://lib.deno.dev/x/abc@~1.3/mod.ts"
import abc from "https://lib.deno.dev/x/abc@>=1.2/mod.ts"
import abc from "https://lib.deno.dev/x/abc@^1.2/mod.ts"
import abc from "https://lib.deno.dev/x/abc@1/mod.ts"
// weak semver style
import abc from "https://lib.deno.dev/x/abc@v1/mod.ts"
import abc from "https://lib.deno.dev/x/abc@v1.3/mod.ts"
import abc from "https://lib.deno.dev/x/abc@v1.3.3/mod.ts"
// ↓
import abc from "https://deno.land/x/abc@v1.3.3/mode.ts"
```

## Motivation

In the modern CDN providers such as jsDelivr and unpkg, we can use the __version alias__ to retrieve a patch update without modifications. We bring this system to deno.land using Deno Deploy. You can get the latest compatible version every cache updates.

## Remaining Problem

`lib.deno.dev` is running on Deno Deploy. If you want to use this domain, I will unlink it soon.
However, I prefer `pkg.deno.dev` like `pkg.go.dev` but it was taken. So, I want to ask you what domain is suitable for.

0. `lib.deno.land`, `lib.deno.dev`
1. `pkg.deno.land`, `pkg.deno.dev`
2. `semver.deno.land`, `semver.deno.land`
3. `alias.deno.land`, `alias.deno.dev`


lucacasonato:
My main concern with this is that versions for modules on deno.land/x do not have to follow semver. they can use any versioning scheme they want

tani:
@lucacasonato Please see [this line](https://github.com/tani/deno_website2/blob/cdbd5e089fd45ab3584b403c234efbede4784164/semver/mod.ts#L48). If the tag (e.g., `v1.2.3`) and range (e.g., `v1.x`) are not compatible with semver spec, then the service fallback.

```typescript
import abc from "https://lib.deno.dev/abc@non-semver/mod.ts"
// ↓
import abc from "https://deno.land/abc@non-semver/mod.ts"
```

tani:
Dear @ry @lucacasonato @wperron,

Thank you for giving comments on GitHub and Discord.
I have written a more formal proposal, which I hope you will accept.

## Background
One of the strong points of Deno compared to Node.js is the compliance of web standards.
The interpreter loads ES modules with an `import` statement, which takes a URL string
and a path string to a local file. Internally, the interpreter stores the ES modules in only one system cache directory.
Hence, the interpreter reduces the occupancy of a local disk.
Moreover, this is a useful feature for single-file ad-hoc scripts because the interpreter downloads the script automatically.

```ts
import abc from "https://deno.land/x/abc@1.3.3/mod.ts"
```

## Problem of Import Statement
Some member of deno community claims two problems about es modules.

1. Deno downloads all the versions of a module even if the version is different from a patch number only.
   Hence, a dependency tree commonly becomes a large tree. For instance, module A depends on modules B and C,
   which depend on a modules D, versions of which are 1.3.3 and 1.3.2, respectively,
   then deno downloads both modules `D@1.3.3` and `D@1.3.2`.
   Thus, the dependency tree becomes larger and larger.
2. We have to update version numbers for each update manually.
   In a simple case, if the project consists of 10 files that depend on ten external modules
   for each file, and then we have to rewrite version numbers on 100 lines.

## Related Works

### DSR (https://dsr.edjopato.de/)
dsr is the original of this proposal.
The service interprets some semver and redirects to deno.land.
It is written in Rust and runs very fast, but the semver is partially incompatible with package.json and deno.
The paths to the packages are also incompatible with deno.land,
which makes it somewhat difficult for each project to migrate from deno.land.

### Import Maps
Import maps are a good solution that follows web standards.
By writing pointers to dependent libraries in a separate file,
you can use a pre-specified short name in the source code without writing the URL.
Furthermore, to update the version of a module, you can change the version at once by rewriting the import maps.
However, you can't specify the version by semver like NodeJs package.json, and you can't resolve dependency-hell.

### CDNs for NPM packages
Skypack, Esm.sh, and Esm.run are CDN services that convert packages uploaded on npmjs.com to ES Module and deliver them.
These services compile transparently on the edge network.
In addition, they compile and interpret the version directive specified in the URL
and automatically deliver the most appropriate version of the package.
However, We are not sure if these services will work.
However, these services only deliver packages uploaded to npmjs.com, not necessarily projects registered on deno.land.

### Deno Package Managers

Package managers made by deno, such as deno-udd, fall into two main categories.
One is to prepare an external file, like import-maps, and install modules based on the specified version.
The other is to add `#` to the URL and manage the version based on the directive at the end.
This method is not intended to solve dependency-hell and requires a separate software installation.

## Version Aliasing for Deno.land

### Semver

Semver is a simple set of rules and requirements that dictate how version numbers are assigned and incremented.
It is commonly used in the JavaScript community.
Although we have no statistical information on how many libraries follow this specification in Deno community,
we believe this is a major manner because the Deno community is close to the JavaScript community.
We propose the redirection of requests based on this rule as follows.

1. The client sends HTTP request to the server to retrieve modules.
2. The server searches git tags for the closest version of the library.
3. The server rewrites the URL with the version and return the 302 redirections.
4. The client sends HTTP request to the redirected URL.

Our proposal is to implement step 2 and 3 based on [deno-semver](),
which is compatible with package.json.

### Non-semver

We also concern that some of the libraries do not follow the semver spec.
Hence, the system should fall back to be compatible with the existing system.
In above step 2, if the program could not found the version because the request or the library did not follow the semver,
the server does not rewrite a version number. Thus, the client sends the same request to deno.land.

## Roadmap

The current system is running on Deno Deploy without other external services,
As it is easy for anyone to launch the system, we propose the migration plan as follows.

#### pkg.deno.land

Launch the semver redirection service with `pkg.deno.land` domain.
The version aliasing service redirects the request to `deno.land`

#### (Optional) deno.land and raw.deno.land

As we think this system has the backward compatibility for `deno.land` regarding the version spec,
we move `deno.land` to `raw.deno.land`,
and move `pkg.deno.land` to `deno.land`.
The version aliasing service redirect the request to `raw.deno.land`.

Finally, the service enables version aliasing without the breaking change.

wperron:
Thanks for the writeup! I think for now going forward with following semver resolution on /x is already a good step forward. I'll review the patch more closely tomorrow, but barring any objections from other team members I think we'll be able to move forward with the change.

The idea of moving /x to a different subdomain has come up internally; It's still unclear whether we want to move forward with it or not, or what shape it'll take. In any case, it's certainly beyond the scope of this ticket anyway, but we certainly appreciate the feedback.

tani:
> Thanks for the writeup! I think for now going forward with following semver resolution on /x is already a good step forward. I'll review the patch more closely tomorrow,

Thank you for taking your time to review this request.

>but barring any objections from other team members I think we'll be able to move forward with the change.

I am looking forward to emerging this request! 

>  The idea of moving /x to a different subdomain has come up internally; It's still unclear whether we want to move forward with it or not, or what shape it'll take. 

It seems a large change for the community. I hope it would be shorter than now like `de.no/abc@v1.2.3` ;)
Anyway, please ask me anything about code and test cases.

Best regards,

wperron (changes requested):
Left a couple of small comments for you to check out. Outside of those, I'd say it's probably best to move the logic to be inside of the `util/` directory, under `semver.ts` and `semver_test.ts`.

tani:
@wperron I updated it.
- Added `resolveSemver` function to registry_utils.ts
- Added test of `resolveSemver` to registry_utils.test.ts
- Added one line in registry.ts to resolve semver

I concern the compatibility between node-semver and deno-semver.
It might change the behaviour if this website moves from Cloudflare Workers to Deno Deploy.

tani:
One more thing, we have to take care of the following situation.

There are tags `1.1`, `1.x` in library A. Then the service resolves `https://deno.land/x/A@1.x/mod.ts` as `https://deno.land/x/A@1.1/mod.ts`. This is a breaking change for all users. I propose you to separate services as I first proposed.

> ## Roadmap
> 
> The current system is running on Deno Deploy without other external services,
> As it is easy for anyone to launch the system, we propose the migration plan as follows.
> ### pkg.deno.land
> 
> Launch the semver redirection service with pkg.deno.land domain.
> The version aliasing service redirects the request to deno.land
> ### (Optional) deno.land and raw.deno.land
> 
> As we think this system has the backward compatibility for deno.land regarding the version spec,
> we move deno.land to raw.deno.land,
> and move pkg.deno.land to deno.land.
> The version aliasing service redirect the request to raw.deno.land.
> 
> Finally, the service enables version aliasing without the breaking change.

wperron:
> I concern the compatibility between node-semver and deno-semver.
It might change the behaviour if this website moves from Cloudflare Workers to Deno Deploy.

Do not fear, `deno.land` is already being served from Deploy! :tada: 

As for it being a breaking change, I'll have to check if any modules are in a situation where some of their versions would be impacted by the change but I would assume that would be a minority (again, will check the actual data).

util/registry_utils.ts
```diff
@@ -1,4 +1,5 @@
 /* Copyright 2020 the Deno authors. All rights reserved. MIT license. */
+import semver from "semver";
```
wperron, line 2: As I said in my other comment, the website is served from Deno Deploy, so you can import from `deno.land/x/semver` just as you were in the original version. You also don't have to add `semver` to the package.json because of that.

wperron (changes requested):

util/registry_utils.ts
tani: Okay. I updated it.

tani:
>As for it being a breaking change, I'll have to check if any modules are in a situation where some of their versions would be impacted by the change but I would assume that would be a minority (again, will check the actual data).

I see. Please check it. Thank you.

tani:
@wperron Could you show me a schedule or roadmap of this feature request? I guees you are busy now. It is no problem. I just to know the schedule.

wperron:
Hey @tani sorry, PR just slept in a crack there. I talked with @lucacasonato and we don't really have any objections to merging it right away except for what you also mentioned about it being a potential breaking change. It would also add latency because there's an extra hop to do to get the list of versions.

What we're thinking here is to add a query string parameter `?semver` -- if it's present, then we'll do the semver resolution, if it's not we'll just use the exact version match that we currently have. Current code doesn't break, and we avoid introducing extra latency.

tani:
> Hey @tani sorry, PR just slept in a crack there. I talked with @lucacasonato and we don't really have any objections to merging it right away except for what you also mentioned about it being a potential breaking change. It would also add latency because there's an extra hop to do to get the list of versions.

Thank you for replying me quickly. I am happy to hear that your don't really have any objections to merging it right away.
Your concerns are the latency and the compatibility. I see. 

> What we're thinking here is to add a query string parameter ?semver -- if it's present, then we'll do the semver resolution, if it's not we'll just use the exact version match that we currently have. Current code doesn't break, and we avoid introducing extra latency.

We have to discuss this deeply as the query parameter is not suitable with import maps (https://github.com/WICG/import-maps/issues/134). Let me explain what the problem is. 

## Case 1: `?semver` parameter

We have to write a redundant code for each file in import-maps if we employ query parameter `?semver`.

```js
// import map
{
  "imports": {
     "abc_modts": "https://deno.land/x/abc@1.x/mod.ts?semver",
     "abc_utilts": "https://deno.land/x/abc@1.x/util.ts?semver"
  }
}
// code
import abc from "abc/mod.ts"
import abc from "abc/util.ts"
```

## Case 2: different domain

On the other hand, we can keep the code simple if we use other domain as follows.

```js
// import map
{
  "imports": {
     "abc": "https://pkg.deno.land/x/abc@1.x/",
  }
}
// code
import abc from "abc/mod.ts"
import abc from "abc/util.ts"
```
## Case 3: different domain

Or, we can use different notation as follows.

```js
// import map
{
  "imports": {
     "abc": "https://deno.land/x/abc:1.x/",
  }
}
// code
import abc from "abc/mod.ts"
import abc from "abc/util.ts"
```

## Suggestion

 I am sorry for that to say again and again. Would you mind considering the different domain first, i.e., you use `pkg.deno.land` to semver routing as I said. I believe this way is the easiest way to start the service.

wperron:
As I said earlier, using different subdomains is something we have discussed internally before, but is not something we want to do in the short term because frankly, we don't really have the bandwidth at this time to take on that project just now. For the scope of this PR, we're comfortable moving forward with the query string parameter solution.

lucacasonato:
@tani Thank you for the work on this.

We are currently not planning to land this, so I'll close this PR to reflect that.

This does not mean we never want to land anything like this, it just means we won't do it right now. Your discussion will be useful when we revisit this in the future.

PR #1830, opened 2021-08-18
feat: show Deploy in the header with 1% probability

satyarohith:
This PR also re-positions the install link as the main call to action and moves the version link to the bottom of the install button.

With "Deploy" in the header:
CleanShot 2021-08-18 at 11 33 06@2x

Without "Deploy" in the header:
CleanShot 2021-08-18 at 11 27 41@2x

---

The failing test of the PR is unrelated to the changes.

### DIFF
--- a/components/Header.tsx
+++ b/components/Header.tsx
@@ -1,8 +1,24 @@
 /* Copyright 2020 the Deno authors. All rights reserved. MIT license. */
 
-import React, { useState } from "react";
+import React, { useEffect, useState } from "react";
 import Link from "next/link";
 import Transition from "./Transition";
+
+/**
+ * Indicates if the "Deploy" link in the header
+ * should be displayed to the user.
+ */
+function displayDeploy() {
+  const showDeploy = localStorage.getItem("showDeploy");
+  if (showDeploy !== null) {
+    return showDeploy === "true";
+  } else {
+    // 1% probability
+    const showDeploy = Math.random() < 1 / 100;
+    localStorage.setItem("showDeploy", String(showDeploy));
+    return showDeploy;
+  }
+}
 
 function Header({
   subtitle,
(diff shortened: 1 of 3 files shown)


components/Header.tsx
```diff
@@ -1,9 +1,25 @@
...
+ * should be displayed to the user.
+ */
+function displayDeploy() {
+  const showDeploy = localStorage.getItem("showDeploy");
+  if (showDeploy !== null) {
+    return showDeploy === "true";
+  } else {
+    // 1% probability
+    const showDeploy = Math.random() < 1 / 100;
+    localStorage.setItem("showDeploy", String(showDeploy));
```
ry, line 18: it seems like over time, if people keep visiting the site they will eventually be pushed into the "showDeploy" category?

satyarohith: The else block should only run on the first visit unless the "showDeploy" key is cleared from localStorage.

components/Header.tsx
```diff
@@ -12,6 +28,8 @@ function Header({
   widerContent?: boolean;
 }): React.ReactElement {
   const [menuOpen, setMenuOpen] = useState(false);
+  const [showDeploy, setShowDeploy] = useState(false);
+  useEffect(() => setShowDeploy(displayDeploy()), []);
```
kt3k, line 32: Seems like it's simpler to use `useMemo`

```suggestion
  const showDeploy = useMemo(() => displayDeploy(), []);
```

satyarohith: `useMemo()` will also run on the server.

kt3k: Ah, ok. makes sense!

ry (approved):
LGTM - assuming you can get CI green

PR #1833, opened 2021-08-20
chore: further optimize SVGs

alrra:
The changes are lossy, however, they should not be noticeable to the human eye.

| Image | Size before | Size after | Size Reduction | Machine image diff |
|:---|:---|:---|:---|:---|
|  `logo.svg` |   6020 B |   4596 B | 1424 B (23.65%) |  |
|  `Deno_Web_Sockets.svg` |   9658 B |      4518 B | 5140 B (53.22%) |  |


kt3k (approved):
LGTM. Thanks!

BTW we switched the main logo to the different one in #1836. So we're accepting the change of `Deno_Web_Sockets.svg`.

PR #1835, opened 2021-08-24
Register .mjs as a JavaScript extension

petamoriken:
`.mjs` is already widely used as an extension for JavaScript.
I want to register `.mjs` for syntax highlighting.


lucacasonato (commented):
Can you also add `.cjs` while you are at it?

petamoriken:
@lucacasonato done!

PR #1838, opened 2021-08-28
Fixed link to std repository

ayame113:
I think the standard library repository has moved to deno_std. This pr fixes that link.


satyarohith (approved):
Thanks

PR #1844, opened 2021-09-02
1841 remove duplicated deno text

Codercise:


kt3k:
In mobile version there is no deno log anymore. This looks strange to me.

スクリーンショット 2021-09-03 15 04 34

In 3rd party module page (`/x/`), the header position looks moved left, and this change also removes the sub page title (in this case `Third Party Modules`). I think that's not desirable

スクリーンショット 2021-09-03 15 06 22

ry:
Thanks for the patch @Codercise, but I'm going to do this slightly differently in https://github.com/denoland/deno_website2/pull/1846

PR #1845, opened 2021-09-05
Use unavatar for Twitter profile images

kidonng:
This is a small improvement to the Artwork, using [unavatar](https://unavatar.io/) for Twitter profile images instead of unstable `twimg.com` links.

No one can ensure unavatar is always working, but it's open source and still better than having broken avatar right now.

### DIFF
--- a/artwork.json
+++ b/artwork.json
@@ -98,7 +98,7 @@
     "alt": "the deno mascot standing between a some tall lit up buildings at night",
     "artist": {
       "name": "Dimitrij Agal",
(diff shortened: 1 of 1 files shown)


satyarohith (approved):
LGTM. Thanks, @kidonng!

kidonng:
Hmm, I literally just changed my mind and decide to use GitHub avatars to replace them instead... but if the author is putting Twitter avatars here, it is probably what they prefer, so I guess it's fine 🤷‍♂️

PR #1846, opened 2021-09-06
Remove double 'Deno' text on index page

### LINKED ISSUE #1841: "Deno" is written twice above the fold on the main screen
Screen Shot 2021-09-01 at 8 55 45 AM

Maybe remove the "Deno" text in the upper left?

ry:
Fixes #1841

### DIFF
--- a/components/Header.tsx
+++ b/components/Header.tsx
@@ -7,9 +7,11 @@ import Transition from "./Transition";
 function Header({
   subtitle,
   widerContent,
+  main,
 }: {
   subtitle?: string;
   widerContent?: boolean;
+  main?: boolean;
 }): React.ReactElement {
   const [menuOpen, setMenuOpen] = useState(false);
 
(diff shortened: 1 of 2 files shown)


components/Header.tsx
```diff
@@ -24,9 +26,11 @@ function Header({
...
             <img className="h-10 w-auto sm:h-12 my-2" src="/logo.svg" alt="" />
             <div className="ml-5 flex flex-col justify-center">
-              <div className="font-bold text-gray-900 leading-tight text-2xl sm:text-3xl tracking-tight">
-                Deno
-              </div>
+              {!main && (
+                <div className="font-bold text-gray-900 leading-tight text-2xl sm:text-3xl tracking-tight">
+                  Deno
+                </div>
+              )}
```
satyarohith, line 33:
```suggestion
```

Can we remove it instead of using a flag?

ry: I want it to say Deno on pages other than index.

satyarohith (approved):
LGTM.

Issue #1879, opened 2021-09-15
Manual search provides duplicate results


kidonng:
It seems DocSearch is indexing duplicate results, result for `v1.11.3` will show up before the one pointing to the latest version.

![image](https://user-images.githubusercontent.com/44045911/133366163-f446f6b4-1921-48ee-b0ea-ef42d0e3a355.png)

![image](https://user-images.githubusercontent.com/44045911/133366202-ead6efcf-cfa1-4019-b407-454489f2fbf1.png)

Also due to #1801 docs for versions before `v1.12.1` doesn't provide a working edit link.

kidonng:
Well, looks like the issue has been resolved.

PR #1880, opened 2021-09-15
Fix edit link for old versions of manual

kidonng:
See #1879


kt3k (approved):
@kidonng LGTM. Thanks!

PR #1884, opened 2021-09-16
add new 2021 logo

nnmrts:
I don't know the actual creation date and didn't want to assume the license so I added "UNKNOWN" there and made this a draft PR for now. @hashrock might have to look over this.

Also the logo referenced here is actually the SVG version created by @jaydenseric here: https://github.com/denoland/dotland/pull/1836. Don't know if they should be added to this JSON as well.

Sorry for pinging you guys and I hope I did everything correctly. 😶‍🌫️

### DIFF
--- a/artwork.json
+++ b/artwork.json
@@ -244,5 +244,19 @@
       "web": "https://pbu.cl/"
     },
     "license": "MIT"
+  },
+  {
+    "date": "2021-08-26",
+    "image": "/logo.svg",
+    "title": "New Deno logo",
+    "alt": "The new official logo of Deno",
+    "artist": {
+      "name": "hashrock",
+      "profile_image": "https://avatars0.githubusercontent.com/u/3132889",
+      "twitter": "hashedrock",
+      "github": "hashrock",
+      "web": "https://hashrock.studio.design/"
+    },
+    "license": "UNKNOWN"
   }
 ]


hashrock:
This is a graphic I made in 2020. The logo is often changed to match the top page design, so this may not be treated as a "new logo for 2021".

I think it's a good idea to make it MIT, but what do you think? @ry

jaydenseric (changes requested):
@nnmrts thanks for tagging me, you're right that there are some nuances here about authorship and licensing.

This metadata reads like the `/logo.svg` asset was created by @hashrock when really they only did the raster artwork. I created the SVG being referenced from scratch; it took hours of time and expertise to optimize the code by hand. Over time, other `denoland/dotland` repo contributors that work on this SVG will also be unacknowledged.

I'm not a lawyer, but there could be a lot of problems associated with the main Deno logo being MIT licensed by the artist. To fulfill the obligations of the MIT license, would we need to include the license in the source code of the SVG (increasing download size)? Or outside the file, in the main repo [LICENSE](https://github.com/denoland/dotland/blob/main/LICENSE) file? The license would need to be included in every repo/project that uses the logo? I already started to use my SVG representation of the original artwork on my own website in the [“My stack” section](https://jaydenseric.com/#my-stack) some time ago. Would that not be a violation of the MIT license, but a violation of using something that was unlicensed since MIT is only being publicly declared now? I thought it was ok to use as it was what Deno was using as the main logo at [https://deno.com](https://deno.com). I also used a raster conversion of my SVG as the icon for the [Deno Nova extension](https://extensions.panic.com/extensions/jaydenseric/jaydenseric.deno/).

I think the authorship metadata here in `artwork.json` is really challenging to get right. It needs to contain two parts; the design authors, and particularly in the case of SVG, the actual asset authors. Sometimes they are the same person, but sometimes they are not and both are important to acknowledge. Sometimes either can be multiple authors.

At first I wondered if the asset is in a GitHub repo like this one is we could somehow link to a commit history for the file, that contains all the details about all the code authors. Unfortunately this is not so easy, because while GitHub allows you to see a commit history for a given file at a particular path (e.g. [here](https://github.com/denoland/dotland/commits/main/public/logo.svg)), if the file is moved to a different path that link could be broken and the old contributors would no longer be acknowledged if you update the link. Also, if an existing asset for a totally unrelated artwork was replaced at the same path (as is the case for `public/logo.svg`), the contributors for both assets will be listed together. Unfortunately GitHub doesn't have a feature that allows filtering away commits older than a particular commit when the asset was replaced.

nnmrts:
While researching another thing today, incidentally I once again stumbled upon https://choosealicense.com/, which recommends several [non-software licenses](https://choosealicense.com/non-software/).

It may be worth looking at how for example [npm](https://docs.npmjs.com/policies/trademark) or the [OpenJS Foundation](https://openjsf.org/wp-content/uploads/sites/84/2021/01/OpenJS-Foundation-Trademark-Policy-2021-01-12.docx.pdf) handles that stuff, but also, these are full on registered trademarks so I don't know if that really applies here.

hashrock:
Hi @jaydenseric, thanks for making the SVG version. I'm sorry if my earlier comment seemed to ignore your contribution.
Of course, I'm assuming that this file contains the creative work of you and other contributors.

It's complicated to have multiple authors. Would it be more convenient to not dare to clarify the license?
Maybe it would be better to leave it UNKNOWN and add the names alrra, jaydenseric, koji. We'll have to add more contributors as we go along.

nnmrts:
Sorry for referencing that site again (😂) but not choosing a license at all [doesn't seem like a good idea](https://choosealicense.com/no-permission/). Ultimately it's the decision of you guys but leaving it as "UNKNOW" or "No license" is the worst-case outcome I can think of.

kt3k:
> To fulfill the obligations of the MIT license, would we need to include the license in the source code of the SVG

I feel this is not expected for logos. So I guess the right choice would be something like `CC0-1.0`.

@ry What do you think?

ry:
Closing because stale

PR #1890, opened 2021-09-23
fix: handle connection closed error

kt3k:
`respondWith` returns a promise (though [typed as void](https://deno.land/x/deploy@0.4.0/types/deploy.fetchevent.d.ts#L10)), and that promise is rejected when the client closed the connection before response finished. And because we don't handle that rejection, that situation causes the crash of the isolate.

### DIFF
--- a/worker/main.ts
+++ b/worker/main.ts
@@ -4,6 +4,10 @@ import { handleRequest, withLog } from "./handler.ts";
 
 const handler = withLog(handleRequest);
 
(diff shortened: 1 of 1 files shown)


lucacasonato (commented):
Consider using std/http instead. It handles all these intricacies.

kt3k:
Confirmed (with Deno CLI) that `std/http` also solves the same issue.

lucacasonato (approved):

PR #1893, opened 2021-09-27
fix: substitute correct std version to $STD_VERSION

### LINKED ISSUE #1787: Substituted std version does not change when changing std version drop down
1. Go to https://deno.land/std@0.99.0/flags
2. Change std version in drop down to anything except 0.99.0

After step 2, the substituted std version will not match the version selected.

kt3k:
Fixes #1787

This PR fixes the substitution of $STD_VERSION in registry rendering.

This PR also adds test cases of path parsing utilities of registry.

### DIFF
--- a/components/Registry.tsx
+++ b/components/Registry.tsx
@@ -23,6 +23,7 @@ import {
   getVersionDeps,
   listExternalDependencies,
   getBasePath,
+  parseQuery,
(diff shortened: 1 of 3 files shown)


satyarohith (approved):
Nice.

LGTM.

PR #1898, opened 2021-09-28
chore: update next and webpack

kt3k:
This PR updates next and webpack.

supercedes #1843


kt3k:
This seems causing a lot of type errors during build 🤔

PR #1902, opened 2021-10-02
feat(registry): Add on the fly compilation

### LINKED ISSUE #1728: Serve precompiled JavaScript to the browser
## Reason

It would be nice to be able to directly import TypeScript modules from the browser without having to use `deno bundle`.

## Suggestions

* Every TypeScript should be compiled once to JavaScript.
* When requested `Accept` is `*/*`, serve the compiled JavaScript. Only serve TypeScript files when `Accept` contains `application/typescript` or when `User-Agent` is `Deno/<VERSION>`.

AnInternetTroll:
fixes #1728


kt3k:
I'll do some performance check for this change on Deploy. `deno.land/x/swc` has 5+MB wasm dependency and that would impact the performance of the start up time of the worker

worker/registry.ts
```diff
@@ -77,22 +81,45 @@ export async function handleRegistryRequest(url: URL): Promise<Response> {
...
+      !resp2.headers.get("content-type")?.includes("javascript")
+    ) {
+      resp2.headers.set("content-type", "application/javascript");
+    } else if (
+      remoteUrl.endsWith(".tsx") &&
+      !resp2.headers.get("content-type")?.includes("typescript")
+    ) {
+      resp2.headers.set("content-type", "application/typescript");
+    }
+  } else if (headers.get("accept")?.includes("javascript")) {
```
kt3k, line 101: Maybe we should also include the case of `Accept: */*`?

AnInternetTroll: I decided against this because if you aren't requesting anything specific you would expect to get the raw resource, without it being changed/transpiled, though I'm fine either way to support older systems that might not send the appropriate header.

AnInternetTroll:
> Regarding the design aspect, this change responds with javascript context for the `path/to/foo.ts`. That would work for the browsers, but might be confusing to the users.

IMO as long as `import { bold } from "https://deno.land/std@0.110.0/fmt/colors.ts"` works then it doesn't matter that much to the user. If they're in an environment where TypeScript is important (like deno) then it will do all the background magic of adding types, but in theory it shouldn't change the behavior of the code.

kt3k:
This seems to slow down the start up time of the isolate from about 1.5 sec to 5.0 sec from my location (my location already has 1.0 sec penalty on start up BTW).

I'm not sure how much impact it does have to the service. Can't predict for sure whether it's safe to merge this or not (should be merged very carefully anyway)

Some numbers to share: deno.land has about 66,000 reqs/hour at the peak today. It's 18.3 reqs/sec. 3.5 sec penalty means additional 64.2 reqs are on hold during deployment (if I understand the deployment process correctly). These numbers are for the entire regions, not a single region. (Deno Deploy has 25 regions, but I guess loads wouldn't be distributed evenly)

kt3k:
Regarding the functional aspect, I was able to use some of std/collections functions in the browser. The compilation itself seems working! 👍 

An example I was able to run on a browser:
```html
<script type="module">
  import { sortBy } from "https://dry-horse-89.deno.dev/std@0.110.0/collections/mod.ts";

  const people = [ 
    { name: "Anna", age: 34 },
    { name: "Kim", age: 42 },
    { name: "John", age: 23 },
  ];  
  const sortedByAge = sortBy(people, (it) => it.age);
  console.log(sortedByAge);
</script>
```
 (I deployed this branch at https://dry-horse-89.deno.dev)

UltiRequiem:
I really would like this feature, is there something I can do to help?

crowlKats:
Ideally this would not be on the fly, but rather when a tag is created

AnInternetTroll:
It's a part of the [roadmap](https://github.com/denoland/deno_registry2/issues/35) on the backend. So you should look in that repo first, then we can either change this PR to use that saved stripped version or make a new PR

josephrocca:
@crowlKats Could/should this perhaps be done lazily? So we keep this pull request, but then add another which optimises it by saving the transpiled JS code to `$NAME/versions/$VERSION/stripped/$FILEPATH.json`, and of course tries to serve from there before transpilation?

ayame113:
I tried deploying this and it looks like the isolate start time has improved to 3.10ms.
Since there are no performance issues anymore, isn't it okay to merge this?

image

worker/registry.ts
```diff
@@ -88,22 +90,48 @@ export async function handleRegistryRequest(url: URL): Promise<Response> {
...
-    remoteUrl.endsWith(".jsx") &&
-    !resp2.headers.get("content-type")?.includes("javascript")
+    headers.get("accept")?.includes("application/typescript") ||
+    headers.get("user-agent")?.includes("Deno/")
   ) {
-    resp2.headers.set("content-type", "application/javascript");
+    // JSX and TSX content type fix
+    if (
+      remoteUrl.endsWith(".jsx") &&
+      !resp2.headers.get("content-type")?.includes("javascript")
```
KSXGitHub, line 101: It seems that `resp2.headers.get("content-type")` is repeated multiple times. Why don't you store it in a variable?

worker/registry.ts
```diff
@@ -88,22 +90,48 @@ export async function handleRegistryRequest(url: URL): Promise<Response> {
...
+    } else if (
+      remoteUrl.endsWith(".tsx") &&
+      !resp2.headers.get("content-type")?.includes("typescript")
+    ) {
+      resp2.headers.set("content-type", "application/typescript");
+    }
   } else if (
-    remoteUrl.endsWith(".tsx") &&
-    !resp2.headers.get("content-type")?.includes("typescript")
+    headers.get("accept")?.includes("javascript") ||
```
KSXGitHub, line 111: `headers.get("accept")` is also repeated.

PR #1915, opened 2021-10-19
feat: server-side analytics

satyarohith:


piscisaureus:
@satyarohith We also have to add user_agent to the reported events.

PR #1916, opened 2021-10-20
Server side analytics - UA endpoint

piscisaureus:
Currently deployed to [dotland.deno-staging.dev](https://dotland.deno-staging.dev/), Google Analytics dashboard is [here](https://analytics.google.com/analytics/web/#/realtime/rt-overview/a210197595w290522871p253551069/).

I realize this makes use of the "legacy" google analytics. 
I'm not suggesting that it is better than what @satyarohith proposed in #1915.

But it seems that GA4 is entirely focused on ecommerce. 
#1915 is sending page views to it but GA4 doesn't seem to do anything useful with the information.

### DIFF
--- a/worker/analytics.ts
+++ b/worker/analytics.ts
@@ -1,9 +1,12 @@
 import { delay } from "https://deno.land/std@0.112.0/async/mod.ts";
 import { ConnInfo } from "https://deno.land/std@0.112.0/http/server.ts";
 
-const GA_BATCH_ENDPOINT = "https://www.google-analytics.com/batch";
 const GA_TRACKING_ID = Deno.env.get("GA_TRACKING_ID")!;
+if (!GA_TRACKING_ID) {
+  console.log("GA_TRACKING_ID not set, Google Analytics reporting disabled.");
+}
 
+const GA_BATCH_ENDPOINT = "https://www.google-analytics.com/batch";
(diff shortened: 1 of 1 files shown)


piscisaureus:
> We need to not track pageview events for things like https://deno.land/_next/static/chunks/main-759ceeaa5839f129d853.js.

Do you have a suggestion on how we can filter those out?

satyarohith:
> > We need to not track pageview events for things like https://deno.land/_next/static/chunks/main-759ceeaa5839f129d853.js.
> 
> Do you have a suggestion on how we can filter those out?

We can have an array of paths and ignore processing a request if its path starts with any of the prefixes in the array. I'll push the change.

lucacasonato:
> We can have an array of paths and ignore processing a request if its path starts with any of the prefixes in the array. I'll push the change.

I suggest we do the reverse. We only track requests if the content type is either text/html, or the path starts with /x

piscisaureus:
> > We can have an array of paths and ignore processing a request if its path starts with any of the prefixes in the array. I'll push the change.
> 
> I suggest we do the reverse. We only track requests if the content type is either text/html, or the path starts with /x

Don't forget /std.* 

Side note, I noticed that deno.land responds with 200 when a page is not found. Intentional?

PR #1923, opened 2021-10-26
Improve HTTP benchmark charts

piscisaureus:
cc @lucacasonato


ry:
Screen Shot 2021-11-05 at 11 38 24 AM


@piscisaureus This patch seems to create very strange/wrong results. There is no discussion or comments on the algorithm you're using here...

PR #1942, opened 2021-11-16
Upgrade to Next.js 12

kidonng:
Resolve #1843
Previous attempt #1898

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,11 +21,11 @@
     "date-fns": "^2.22.1",
     "dompurify": "^2.2.6",
     "marked": "^2.0.1",
-    "next": "^10.0.7",
-    "next-plugin-preact": "^3.0.3",
+    "next": "11",
+    "next-plugin-preact": "^3.0.6",
     "orga": "^2.4.8",
-    "preact": "^10.5.12",
-    "preact-render-to-string": "^5.1.12",
+    "preact": "^10.5.15",
+    "preact-render-to-string": "^5.1.19",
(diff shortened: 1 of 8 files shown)


kidonng:
Hi @CLAassistant, nice to [meet](https://github.com/denoland/dotland/pull/1369#issuecomment-679021210) [you](https://github.com/denoland/dotland/pull/1689#issuecomment-789595086) again 🙂

preact.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module "preact/compat/server" {
```
kidonng, line 1: The `@preact/compat` alias is doing a bad job here since Next.js now uses ESM imports, so I just made those imports use `preact/compat` directly. I know it defeats the meaning of `@preact/compat` somehow but we _are_ using Preact anyway so it's not a big deal.

package.json
```diff
@@ -57,8 +56,8 @@
     "eslint-plugin-react": "^7.22.0",
     "isomorphic-unfetch": "^3.1.0",
     "jest": "^26.6.3",
-    "postcss-import": "^14.0.0",
+    "scheduler": "^0.20.2",
```
kt3k, line 59: Is this dependency necessary?

kidonng: Yep: https://github.com/preactjs/next-plugin-preact/issues/49

kt3k (approved):
LGTM. Thank you for your contribution! @kidonng

PR #1943, opened 2021-11-16
chore: update complex example on the homepage

meszarosdezso:
This updates the server example on the home page to use `serve` instead of `listenAndServe`.


satyarohith (approved):
LGTM. Thanks

PR #1949, opened 2021-11-22
Cache proxied requests to now.sh in memory

piscisaureus:
Hopefully this would reduce deno.land downtime.
All static assets combined currently take up about ~30mb so the lack of cache eviction should not cause any immediate problems.
Nonetheless, this is a POC, feel free to criticize.

### DIFF
--- a/worker/handler.ts
+++ b/worker/handler.ts
@@ -95,26 +95,45 @@ export function extractAltLineNumberReference(
   };
 }
 
+const cache = new Map<string, { response: Response; body: ArrayBuffer }>();
+
 async function proxyFile(
   url: URL,
   remoteUrl: string,
   request: Request,
 ): Promise<Response> {
-  const init = {
-    method: request.method,
-    headers: request.headers,
-  };
-  const urlR = remoteUrl + url.pathname;
-  const modifiedRequest = new Request(urlR, init);
-  let lastErr;
+  const proxyUrl = new URL(remoteUrl + url.pathname).href;
+
+  const cacheKey = [request.method, proxyUrl].join(",");
+  let cacheData = cache.get(cacheKey);
+
+  if (cacheData === undefined) {
+    const proxyRequest = new Request(proxyUrl, { method: request.method });
+    const proxyResponse = await fetchWithRetry(proxyRequest);
+    if (!(proxyResponse.ok || proxyResponse.redirected)) {
+      return proxyResponse;
+    }
+
+    cacheData = {
+      body: await proxyResponse.arrayBuffer(),
+      response: proxyResponse,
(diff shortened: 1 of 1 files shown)


lucacasonato (commented):
Could you change this to not cache HTML files?

piscisaureus:
> Could you change this to not cache HTML files?

I can if course, but why?
I’m afraid with that change it wouldn’t help much with deno.land uptime.

lucacasonato:
@piscisaureus Because otherwise we might serve stale pages. All the assets contain content hashes in their names, so they are fine to cache, but the html pages do not contain content hashes.

The issue is that the Deploy worker gets deployed about a minute before the Vercel deployment for a given commit, so a worker would be caching the wrong thing for about a minute. We could also just set a 60 second expiration on HTML pages?

piscisaureus:
@lucacasonato
As discussed offline, it will now flush the cache after 2 min.
This PR also makes it so that etags for non-immutable resources are served correctly so fonts etc are properly cached by the brower.

worker/handler.ts
```diff
@@ -95,26 +99,91 @@ export function extractAltLineNumberReference(
...
+        break;
+      default:
+        throw new Error(`Unsupported request method: ${request.method}`);
+    }
+    return new Response(body, {
+      headers: {
+        "content-type": cacheEntry.contentType,
+        "cache-control": cacheEntry.immutable ? "public,immutable" : "private",
+        "etag": cacheEntry.etag,
+      },
```
lucacasonato, line 170: Why only these 3 headers?

piscisaureus: Because they're the only ones that matter.

Note that I do not forward any request headers to Vercel either, because they might (in theory, not in practice) change the response, which would make the cache too simplistic.

Which ones would you like to see included?

lucacasonato: Seems fine.

worker/handler.ts
```diff
@@ -95,26 +99,91 @@ export function extractAltLineNumberReference(
...
+      case "GET":
+        body = cacheEntry.body;
+        break;
+      default:
+        throw new Error(`Unsupported request method: ${request.method}`);
+    }
+    return new Response(body, {
+      headers: {
+        "content-type": cacheEntry.contentType,
+        "cache-control": cacheEntry.immutable ? "public,immutable" : "private",
```
lucacasonato, line 168:
```suggestion
        "cache-control": cacheEntry.immutable ? "public,max-age=31536000,immutable" : "public,max-age=0,must-revalidate",
```

lucacasonato (approved):

Issue #1964, opened 2021-12-11
[Suggestion] Show deprecation warning on registry's page


meszarosdezso:
## Motivation

The idea came from the standard lib's `ws` module which has been deprecated in favor of the built in websocket API, but it is not very prominent on its registry page on the website. This way the users can easily slip over the fact that it should not be used, and jump directly to the example.

## Possible solution

I thought the page could have some sort of _Deprecation banner_ on the top of the page, telling the user immadiately that they should probably not use the given module / use something else. I attached a quick design about what I had in mind.

An issue with this is that (like the std library) a registry can contain multiple modules inside them, so afaik this couldn't be solved with some metadata.

I thought of two possible solutions at first:

 - include a file named `DEPRECATED` in the same directory as the module's `README` is located, containing the information about what other API or module should be used
 - have a "jsdoc like" comment at the top of the README like `<!-- @deprecated Use the Deno.upgradeWebsocket() API instead -->`

Thoughts are welcome, I would gladly work on it if this is something desired.

![image](https://user-images.githubusercontent.com/45368713/145694516-dac60dbe-845c-4ecd-a0d9-b560a8d3196c.png)

kt3k:
I like the idea. I prefer `<!-- @deprecated ... -->` directive in README to `DEPRECATED` file

I think this feature is also useful for 3rd party modules

meszarosdezso:
Thanks for the response @kt3k, I have opened #1965. If that looks good, I can add the comment to the `ws` module's `README` to see it in action before merging.

kt3k:
Note: discussed this with community members in discord and we want to deprecate the past versions as well, which the current suggested design can't provide.

Because the published modules are immutable, it doesn't work to get the deprecation info from the source code. We need to get the input via some other source to deprecate past versions. We probably need to change denoland/deno_registry2 to achieve that. https://github.com/denoland/deno_registry2/issues/270

crowlKats:
Closing as implemented through doc view

PR #1965, opened 2021-12-12
feat(registry): add deprecated banner

meszarosdezso:
Ref #1964 

I think there should also be a note somewhere about this, could someone point me to a documentation where this would fit?

### DIFF
--- a/util/registry_utils.test.ts
+++ b/util/registry_utils.test.ts
@@ -1,6 +1,6 @@
 /* eslint-env jest */
 
-/* Copyright 2020 the Deno authors. All rights reserved. MIT license. */
+/* Copyright 2021 the Deno authors. All rights reserved. MIT license. */
 
 import {
   fileNameFromURL,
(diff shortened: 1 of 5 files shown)


meszarosdezso:
An example can be visited here since there is no module currently using this directive: https://deno-website2-lmr0yi8c6-denoland.vercel.app/

( I removed the example from the PR )

util/registry_utils.test.ts
```diff
@@ -238,3 +239,15 @@ test("parseQuery", () => {
     path: "/mod.ts",
   });
 });
+
+test("parseDeprecatedDirective", () => {
```
kt3k, line 243: 👍

kt3k (approved):
LGTM

meszarosdezso:
Hey @kt3k, I think this can be closed for now, there were not much response from the community on the issue and this has many conflicts now.

I would happily get back to work on it when there's a final design decision on the feature!

Closing until then.

PR #1966, opened 2021-12-12
Remove cookie banner

piscisaureus:
We don't actually use any cookies, except to track whether the user has
dismissed the cookie banner.

Additionally, there is no legal requirement to ask for consent (GDPR) or
provide an opt-out (CCPA) to use this type of cookie.


lucacasonato (approved):
We used to have cloudflare analytics on the page, which is what the banner was for.

lucacasonato:
I think that was it at least

piscisaureus:
> We used to have cloudflare analytics on the page, which is what the banner was for.

Hardly "functional cookies", but let's say bygones be bygones :)

lucacasonato:
Actually no, was because _cfuid: https://github.com/denoland/dotland/issues/1429. That has since been removed tough

PR #1976, opened 2021-12-26
:sparkles: Explicitly require process

UltiRequiem:


crowlKats:
whats the reasoning for this?

UltiRequiem:
> whats the reasoning for this?

https://nodejs.org/api/process.html#process

> While it is available as a global, it is recommended to explicitly access it via require or import.

Sorry, forgot to put in the description.

kt3k:
Node.js seems to have started recommending non-global `process` and `Buffer` since v16.x https://github.com/nodejs/node/pull/39043

kt3k (approved):
LGTM

PR #1977, opened 2021-12-28
chore: further optimize `logo.svg`

alrra:
The changes are lossy, however, they should not be noticeable to the human eye.

| Size before | Size after | Size Reduction | Machine image diff |
|:---|:---|:---|:---|
| 3716 B | 1990 B | 1726 B (46.45%) |  |


satyarohith (approved):
Thanks

PR #1978, opened 2021-12-28
Revert "Add `/install.{sh,ps1}` URL aliases"

lucacasonato:
Reverts denoland/dotland#1969

it isn’t working for me and some other users on discord.


kidonng:
Sorry for the troublesome PR. I forgot that `deno.land` runs behind a Cloudflare worker which is more complex than what I thought.

What I expect:
1. Visit `deno.land/install.sh`
2. Vercel returns `308` to `deno.land/x/install/install.sh`
3. (same as visiting `deno.land/x/install/install.sh` directly)

What actually happened:
1. Visit `deno.land/install.sh`
2. Cloudflare worker process it in `handleRequest()`
3. Cloudflare worker process it in `proxyFile()`
4. Cloudflare worker fetches file from `deno-website2.now.sh`
5. Vercel returns `308` to `deno.land/x/install/install.sh`
6. Vercel returns a HTML page since the URL matches the route `/x/[...rest]`
7. Cloudflare worker returns the HTML page, which is unexpected

PR #1969, opened 2021-12-19
Add `/install.{sh,ps1}` URL aliases

kidonng:
Fix #1961

Added `/install.sh -> /x/install/install.sh` and `/install.ps1 -> /x/install/install.ps1` aliases.

Also renamed the [deprecated](https://vercel.link/combining-old-and-new-config) `now.json` to `vercel.json`.

PR #1979, opened 2021-12-28
Add URL aliases to installation scripts

kidonng:
This is another attempt of #1969. I don't have any means to test it but it should work properly.

Closes https://github.com/denoland/dotland/issues/1961


satyarohith (approved):
I tested it locally, and it works as expected.

LGTM. Thanks, @kidonng!

Issue #1974, opened 2021-12-24
Implement registry suggestions v2


kitsonk:
Deno CLI now supports registry suggestions v2, we need to implement them for `deno.land`.

PR #1980, opened 2022-01-04
feat: add support for v2 registry suggestions

kitsonk:
Closes: #1974

This PR:

- servers up the v1 or v2 configuration of the registry suggestions based on the client's support
- preserves the v1 suggestions (`/worker/vscode.ts`).
- adds several endpoints under (`/_api/`) which generates the responses for the v2 configuration
- proxies requests to `api.deno.land` pending full migration to deploy

The features it adds:

- Provides a fuzzy search for package/module names
- Provides documentation/details for packages, versions and paths as they are completed
- Provides completions for paths which reveal "folder by folder"
- Provides a summary of each sub-package/module in `std` to make it easier to explore `std`.

### DIFF
--- a/worker/main.ts
+++ b/worker/main.ts
@@ -6,5 +6,5 @@ import { listenAndServe } from "https://deno.land/std@0.108.0/http/server.ts";
 
 const handler = withLog(handleRequest);
 
-console.log("The server is available at http://localhost:8080");
+console.log("The server is available at http://localhost:8081");
 listenAndServe(":8081", handler);
--- a/worker/main.ts
+++ b/worker/main.ts
@@ -1,3 +1,4 @@
+#!/usr/bin/env -S deno run --allow-read=. --allow-net --allow-env
 /* Copyright 2020 the Deno authors. All rights reserved. MIT license. */
 
 import { handleRequest, withLog } from "./handler.ts";
@@ -6,4 +7,4 @@ import { listenAndServe } from "https://deno.land/std@0.108.0/http/server.ts";
 const handler = withLog(handleRequest);
 
 console.log("The server is available at http://localhost:8080");
-listenAndServe(":8080", handler);
+listenAndServe(":8081", handler);
--- a/worker/suggestions.ts
+++ b/worker/suggestions.ts
@@ -1,9 +1,9 @@
-/* Copyright 2021 the Deno authors. All rights reserved. MIT license. */
+/* Copyright 2021-2022 the Deno authors. All rights reserved. MIT license. */
 
 // @deno-types https://deno.land/x/fuse@v6.4.1/dist/fuse.d.ts
 import { default as Fuse } from "https://deno.land/x/fuse@v6.4.1/dist/fuse.esm.js";
 import { prettyBytes } from "https://deno.land/x/pretty_bytes@v1.0.5/mod.ts";
-import twas from "https://esm.sh/twas";
+import twas from "https://esm.sh/twas@2.1.2";
 
 import { S3_BUCKET } from "./registry.ts";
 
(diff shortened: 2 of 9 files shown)


worker/main.ts
```diff
@@ -6,4 +7,4 @@ import { listenAndServe } from "https://deno.land/std@0.108.0/http/server.ts";
 const handler = withLog(handleRequest);
 
 console.log("The server is available at http://localhost:8080");
-listenAndServe(":8080", handler);
+listenAndServe(":8081", handler);
```
lucacasonato, line 10: Port mismatch

worker/suggestions.ts
```diff
@@ -0,0 +1,540 @@
+/* Copyright 2021 the Deno authors. All rights reserved. MIT license. */
+
+// @deno-types https://deno.land/x/fuse@v6.4.1/dist/fuse.d.ts
+import { default as Fuse } from "https://deno.land/x/fuse@v6.4.1/dist/fuse.esm.js";
+import { prettyBytes } from "https://deno.land/x/pretty_bytes@v1.0.5/mod.ts";
+import twas from "https://esm.sh/twas";
```
lucacasonato, line 6: Not versioned!

worker/suggestions.ts
```diff
@@ -0,0 +1,540 @@
...
+function getMeta(pkg: string, ver: string): MetaJson | undefined {
+  return packageMeta.get(pkg)?.get(ver);
+}
+
+/** When dealing with a path, try to preselect the most logical module for the
+ * user. */
+function getPreselectPath(items: string[]): string | undefined {
+  // preselect anything that appears to be a root file in order
+  const preselect = [
+    "mod.ts",
```
lucacasonato, line 209: main.{js,ts} is pretty common too.

worker/suggestions.ts
```diff
@@ -0,0 +1,540 @@
...
+}
+
+/** Given a package, return its latest version. */
+async function getLatestVersion(pkg: string): Promise<string> {
+  const data = versions.get(pkg) ?? await fetchVersions(pkg);
+  return data.latest;
+}
+
+function getMeta(pkg: string, ver: string): MetaJson | undefined {
+  return packageMeta.get(pkg)?.get(ver);
```
lucacasonato, line 201: Is this cache ever invalidated? If I understand right, newly published versions would not show up until the next time this script is restarted.

worker/suggestions_test.ts
```diff
@@ -0,0 +1,160 @@
...
+
+import {
+  assert,
+  assertArrayIncludes,
+  assertEquals,
+  assertStringIncludes,
+} from "https://deno.land/std@0.119.0/testing/asserts.ts";
+
+import { handleApiRequest } from "./suggestions.ts";
+
```
lucacasonato, line 11: It'd be great to have a test for the manifest handler.

lucacasonato (commented):
Looks great already!

worker/suggestions.ts
kitsonk: No, but the `meta.json` is immutable, tied to a specific version, right?  Other things, like what versions there are for a package, are time limited cached, but a `meta.json` is considered immutable.  That is right, correct?

worker/suggestions.ts
kitsonk: 🤦 doh!

worker/suggestions.ts
lucacasonato: Whoops, I commented this on the wrong function. I meant to comment on `fetchPackageData`. The `packages` cache is never invalidated as far as I can tell (or I can't find where it happens, in which case a comment would be nice).

kitsonk: The package data isn't ever invalidated... I considered this a "small" risk, in that the only things that it would impact would be the package description and the star count.  I felt the natural lifecycle of a worker is sufficient.  I have made a note to that effect.

lucacasonato: Ah right, that totally makes sense. `versions` need to be invalidated though because they can change pretty frequently (latest version for example). Maybe just a 5m setInterval that wipes the cache?

lucacasonato (approved):
Only concern is the cache invalidation of the module version list cache. If you think that isn't going to be an issue feel free to not do anything about it.

LGMT, great to have 🚀

worker/suggestions.ts
kitsonk: Yeah, good point about that...  I dislike `setInterval` though... :-) I can do like I do with the package cache and set a "wipe" timestamp and when a request is handled that requires the versions, wipe it before handling the request.

PR #1985, opened 2022-01-05
feat: better content negotiation on code

kitsonk:
This properly fixes a minor improvement that was attempted in #1980 around determining if HTML should be served instead of code to a client.

It also aligns the tests for this to what is occurring in the real world, with an `Accept` header from a recent version of Chromium for browsers and `*/*` for those from the Deno CLI/Deploy which is what is passed from both of them currently.

It also reverts the dependencies served from a previous deployment of the worker.  We need to address how `deno.land` can depend on code itself hosts, but it feels unsafe to continue to leave the temporary URLs in the code base.


kt3k (approved):
LGTM

PR #1988, opened 2022-01-09
add deno simpson artwork & update artist info

SamipPoudel58:
Added a new "Deno Simpson" artwork and updated some info of my previous artworks "Jurassic Deno" and "Glitchy Deno"


ry (approved):
hah - ok thanks

PR #1992, opened 2022-01-18
Use v1 registry config for old Deno versions

lucacasonato:
The v2 registry config was still sent to old Deno versions.


kitsonk (approved):
LGTM 🤦

PR #2015, opened 2022-02-21
feat: add canonical url of manual pages

### LINKED ISSUE #2010: Set canonical url to the latest version of the same page
Google search results sometimes link to the older version of manual. For example, `deno web server` search shows `https://deno.land/manual@v1.14.3/examples/http_server` as the first result on my case

Setting the link tag like `<link rel="canonical" href="url-of-latest-document" />` should solve this issue. ref. https://developers.google.com/search/docs/advanced/crawling/consolidate-duplicate-urls

kt3k:
This adds canonical url meta tags to the manual pages. ref. https://developers.google.com/search/docs/advanced/crawling/consolidate-duplicate-urls

The effect of this change is that the search engines will always list the latest version of the same page in the search result instead of versioned pages.

For example, if I search for the query `deno config` in google now, the search result shows `https://deno.land/manual@v1.16.0/getting_started/configuration_file` (outdated version) for the first result. This will be changed to `https://deno.land/getting_started/configuration_file` (latest version) by this change.

closes #2010

### DIFF
--- a/components/Manual.tsx
+++ b/components/Manual.tsx
@@ -236,6 +236,7 @@ function Manual(): React.ReactElement {
         <title>
           {pageTitle === "" ? "Manual | Deno" : `${pageTitle} | Manual | Deno`}
         </title>
+        <link rel="canonical" href={`https://deno.land/manual${path}`} />
         <link
           rel="preconnect"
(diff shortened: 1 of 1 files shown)


kt3k:
I'll send this again to fresh version of the website later.

PR #2020, opened 2022-02-24
refactor: merge worker into fresh

crowlKats:
Questions:
- [x] which responses do we set `Access-Control-Allow-Origin: *`? all of them?
- [x] what to do with caching?
- [x] What do we do with the current `proxyFile` function?


worker/main.ts
```diff
@@ -1,10 +1,8 @@
...
 
-import { handleRequest, withLog } from "./handler.ts";
+import { handleRequest } from "./handler.ts";
 import { listenAndServe } from "https://deno.land/std@0.108.0/http/server.ts";
 
-const handler = withLog(handleRequest);
-
 console.log("The server is available at http://localhost:8081");
-listenAndServe(":8081", handler);
+listenAndServe(":8081", handleRequest);
```
ry, line 8: `//worker/main.ts` should just be a symlink to `//main.ts` ?

crowlKats: yes, this is just old code to keep reference with while porting

deps.ts
```diff
@@ -1,4 +1,4 @@
-export * from "https://raw.githubusercontent.com/lucacasonato/fresh/04a32f34987fbd59ce646967f216b2cca072a58f/runtime.ts";
+export * from "https://raw.githubusercontent.com/piscisaureus/fresh/d990a0bd6b76b2c2ec4016e9bf2584c7c5c8fe88/runtime.ts";
```
ry, line 1: This is getting a little ridiculous - we need to have a normal review & release process for code going into fresh if we're going to depend on it for our core infrastructure. 

cc @lucacasonato @piscisaureus

ry (commented):
What happens when this PR lands. The dotland.deno.dev still has its entrypoint as `worker/main.ts` but that file is a symlink here? I thought the symlinks where not working?

deps.ts
crowlKats: we moved to bert's fork since luca isnt around to review & merge PRs

crowlKats:
@ry i just committed a workaround

ry:
@crowlKats [The preview deployment](https://dash.deno.com/projects/dotland/deployments/k4016qskg72g) seems broken. Getting this error:
```
TypeError: Must be a file URL.     at fromFileUrl (https://deno.land/std@0.108.0/path/posix.ts:408:15)     at Function.fromRoutes (https://raw.githubusercontent.com/piscisaureus/fresh/d990a0bd6b76b2c2ec4016e9bf2584c7c5c8fe88/src/server/context.ts:120:48)     at https://raw.githubusercontent.com/denoland/dotland/a7f1cd25da1cc471cb4d54d3210e91abadc69478/main.ts:65:33
```

crowlKats:
It really doesn't help that the source mapping is completely fucked up and debugging this is quite annoying as the reported lines dont make sense

completions_v2.ts
```diff
@@ -0,0 +1,198 @@
+// Copyright 2022 the Deno authors. All rights reserved. MIT license.
```
ry, line 1: Is this module new code? I can't tell where it came from?

crowlKats: This is https://github.com/denoland/dotland/blob/main/worker/suggestions.ts, but went through two refactors

server_deps.ts
```diff
@@ -1 +1,10 @@
-export * from "https://raw.githubusercontent.com/lucacasonato/fresh/04a32f34987fbd59ce646967f216b2cca072a58f/server.ts";
+export * from "https://raw.githubusercontent.com/piscisaureus/fresh/d990a0bd6b76b2c2ec4016e9bf2584c7c5c8fe88/server.ts";
```
ry, line 1: Please use https://deno.land/x/deno_fresh@0.0.1/server.ts

deps.ts
ry: Use https://raw.githubusercontent.com/denoland/fresh/0.0.2/runtime.ts

This doesn't include bert's etag patch - we should fix https://github.com/denoland/deploy/issues/2758 before landing this one.

completions_v2.ts
lucacasonato, line 1: Why is this file at the root of the tree?

lucacasonato (commented):
LGTM, if you have your finger ready on the revert button :-)

lucacasonato (approved):

completions_v2.ts
crowlKats: well, it isnt part of pages, static, utils, nor tests, so it has no place to belong, + this is being imported directly by main.ts

PR #2023, opened 2022-02-24
chore: fix error in ci.yml

kt3k:
`web` job in `ci.yml` is now empty, and it causes the validation error below:

スクリーンショット 2022-02-24 19 19 10


crowlKats (approved):
LGTM, my bad, should have noticed that in the PR

kt3k:
I'd say it's github's fault because it doesn't send failure status to PR with yaml errors 😜

PR #2027, opened 2022-02-24
Add link to module-level documentation

surma:
With this PR, the side panel will show a link straight do `doc.deno.land` if the module has a top-level `mod.ts`. If not, the link is hidden.

This implicitly assumes that `mod.ts` is a module’s canonical entry point and I don’t think that’s ever been decided, but it’s better than nothing for now.

Screenshot 2022-02-24 at 14 59 25

Screenshot 2022-02-24 at 14 59 30

### DIFF
--- a/pages/x/module.tsx
+++ b/pages/x/module.tsx
@@ -162,8 +162,12 @@ function ModuleView({
     ? `https://doc.deno.land/https://deno.land${canonicalPath}`
     : null;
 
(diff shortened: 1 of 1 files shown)


roj1512:
@surma Run `deno fmt` to fix formatting.

ry (approved):
LGTM - thanks!

PR #2030, opened 2022-02-25
fix: infinite redirects on non-existing module

crowlKats:


AaronO (approved):
LGTM, undefined !== null

Issue #2031, opened 2022-02-27
Indent based codeblocks cause deployment failure


kt3k:
The url https://deno.land/std@0.127.0/fmt returns `502: Bad Gateway (DEPLOYMENT_FAILED)` error.

Other std submodules don't seem having this issue, for example `std/archive` works (https://deno.land/std@0.127.0/archive)

ref https://github.com/denoland/deno_std/issues/1969

crowlKats:
```
Error rendering page TypeError: Cannot read properties of undefined (reading 'split')
Please report this to https://github.com/markedjs/marked.
    at Renderer.code (https://deno.land/x/gfm@0.1.18/mod.ts:18:25)
    at j.parse (https://cdn.esm.sh/v66/marked@3.0.7/deno/marked.js:48:2211)
    at Function.parse (https://cdn.esm.sh/v66/marked@3.0.7/deno/marked.js:48:1493)
    at p (https://cdn.esm.sh/v66/marked@3.0.7/deno/marked.js:49:2250)
    at render (https://deno.land/x/gfm@0.1.18/mod.ts:45:16)
    at Object.Markdown (file:///Users/crowlkats/projects/denoland/dotland/components/Markdown.tsx:10:18)
    at m (https://cdn.esm.sh/v59/preact-render-to-string@5.1.19/X-ZGVwczpwcmVhY3RAMTAuNS4xNA/deno/preact-render-to-string.js:4:916)
    at m (https://cdn.esm.sh/v59/preact-render-to-string@5.1.19/X-ZGVwczpwcmVhY3RAMTAuNS4xNA/deno/preact-render-to-string.js:9:66)
    at m (https://cdn.esm.sh/v59/preact-render-to-string@5.1.19/X-ZGVwczpwcmVhY3RAMTAuNS4xNA/deno/preact-render-to-string.js:9:66)
    at m (https://cdn.esm.sh/v59/preact-render-to-string@5.1.19/X-ZGVwczpwcmVhY3RAMTAuNS4xNA/deno/preact-render-to-string.js:4:1044)
```

crowlKats:
easy reproducible case:
```ts
import {render} from "https://deno.land/x/gfm@0.1.18/mod.ts";

console.log(render(`
If you are looking for the documentation proper, skip to:

    "printf: prints formatted output"

and
`));
```

crowlKats:
A fix has been merged, keeping this issue open until gfm gets a new release

kt3k:
[...]

PR #2032, opened 2022-02-27
fix: dont fail deployment on indentation based codeblocks

crowlKats:
towards #2031
hotfix until new gfm version gets released


srackham:
Brilliant! All working again.

PR #2035, opened 2022-02-28
Reduce `gfm.css` bundle size

kidonng:
See https://github.com/denoland/dotland/issues/2021#issuecomment-1054403007
The bundle is generated from upstreamed PR: https://github.com/lucacasonato/deno-gfm/pull/2

image

_Look mum ~~no~~ lightweight CSS!_


kidonng:
@crowlKats would you like to review this? The huge bundle is causing too much unnecessary traffic.

crowlKats:
I wonder, should we maybe rather serve the css from gfm directly instead of having a copy ourselves?

kidonng:
Probably, I'm no just following existing conventions.

PR #2040, opened 2022-03-02
Restore inline display of Markdown images

### LINKED ISSUE #2038: Line break between the two images without empty line
Hi,

If there is no empty line between two tags images in Markdown, there should be no line break between the two images.

```Markdown
[![npm][img-npm]][link-npm]
[![jsdelivr][img-jsdelivr]][link-jsdelivr]
[![build][img-build]][link-build]
[![coverage][img-coverage]][link-coverage]
[![semver][img-semver]][link-semver]
```

The markdown above ([full version](https://github.com/regseb/cronnor/blob/v1.3.0/README.md)) is displayed on [Deno](https://deno.land/x/cronnor@v1.3.0/README.md):

screenshot

kidonng:
Fix #2038
Test: https://dotland-fresh-ggcxpmcsd6tg.deno.dev/x/cronnor@v1.3.0/README.md

I'm not sure but probably there are more issues caused by CSS reset 🤷

### DIFF
--- a/static/app.css
+++ b/static/app.css
@@ -183,3 +183,7 @@ ul.tree li:before {
 ul.tree li:last-child:before {
 	border-left: 1px solid rgb(100, 100, 100);
 }
+
+.markdown-body img {
+    display: inline;
+}


crowlKats:
Shouldnt this rather be part of gfm?

kidonng:
> Shouldnt this rather be part of gfm?

I think deno-gfm is general purpose and should not assume other projects are using CSS resets as well.

Take a look at GitHub's style, they don't override `display` of `` as well:

https://github.com/sindresorhus/github-markdown-css/blob/afa572390de862d4bb1cf7ca0c0d4ecf2373ced9/github-markdown.css#L202-L207

kt3k:
twind seems setting `dispaly: block` for img by default. ref: https://github.com/tw-in-js/twind/blob/851160d9895fb2e11d6e6a9911017a5d33c8d348/src/twind/preflight.ts#L246

and we include these settings here: https://github.com/denoland/dotland/blob/a53e5cb029681cb6ae48524fda543b11313f5ff1/pages/_render.ts#L56

these styles are visible, embedded in `<style id="__FRSH_STYLE">` tag in html

kt3k (approved):
LGTM

kidonng:
> twind seems setting `dispaly: block` for img by default. ref: https://github.com/tw-in-js/twind/blob/851160d9895fb2e11d6e6a9911017a5d33c8d348/src/twind/preflight.ts#L246

Right, that's what I mean by "CSS reset".

kt3k:
Ah, ok.

BTW I did some research about reseting `` in this way.

`normalize.css` and `modern-normalize` don't do this reset, but tailwind (and therefore twind) does this https://github.com/tailwindlabs/tailwindcss/blob/f980ca4ca4a74eb39dc1f17f67373f117e2b1052/src/css/preflight.css#L339-L347

This practice seems to have started in the discussion in CSS remedy https://github.com/jensimmons/cssremedy/issues/14

kidonng:
Thanks for the info, that's something new to me.

Block images by default is fine, but this is user content we can't control (hence the `.markdown-body` prerequisite), they should behave like ordinary images and on e.g. GitHub.

PR #2055, opened 2022-03-09
Fix missing whitespace in benchmarks page

gimenete:
There's a missing whitespace as you can see in this screenshot

image

I followed the same approach than in the line above to fix it.


crowlKats (approved):
LGTM, thanks!

PR #2061, opened 2022-03-11
fix: filesize cap on highlighting for modules

crowlKats:


kt3k (approved):
Checked `Max display filesize exceeded` result here: https://dotland-fresh-dhkpcd64yqk0.deno.dev/x/swc@0.1.4/swc_wasm/wasm.js

LGTM

PR #2065, opened 2022-03-15
Resolve semver if ?semver exists

tani:
This pull request is a continuation of #1826 .
This proposal resolves the semver query if `?semver` exists in the request.
This approach is safer than the previous proposal.

Thank you so much for the advice @kt3k 

> ## Background
> 
> One of the strong points of Deno compared to Node.js is the compliance of web standards. The interpreter loads ES modules with an `import` statement, which takes a URL string and a path string to a local file. Internally, the interpreter stores the ES modules in only one system cache directory. Hence, the interpreter reduces the occupancy of a local disk. Moreover, this is a useful feature for single-file ad-hoc scripts because the interpreter downloads the script automatically.
> 
> ```ts
> import abc from "https://deno.land/x/abc@1.3.3/mod.ts"
> ```
> 
> ## Problem of Import Statement
> 
> Some member of deno community claims two problems about es modules.
> 
>     1. Deno downloads all the versions of a module even if the version is different from a patch number only.
>        Hence, a dependency tree commonly becomes a large tree. For instance, module A depends on modules B and C,
>        which depend on a modules D, versions of which are 1.3.3 and 1.3.2, respectively,
>        then deno downloads both modules `D@1.3.3` and `D@1.3.2`.
>        Thus, the dependency tree becomes larger and larger.
> 
>     2. We have to update version numbers for each update manually.
>        In a simple case, if the project consists of 10 files that depend on ten external modules
>        for each file, and then we have to rewrite version numbers on 100 lines.
> 
> 
> ## Related Works
> ### DSR ([dsr.edjopato.de](https://dsr.edjopato.de/))
> 
> dsr is the original of this proposal. The service interprets some semver and redirects to deno.land. It is written in Rust and runs very fast, but the semver is partially incompatible with package.json and deno. The paths to the packages are also incompatible with deno.land, which makes it somewhat difficult for each project to migrate from deno.land.
> ### Import Maps
> 
> Import maps are a good solution that follows web standards. By writing pointers to dependent libraries in a separate file, you can use a pre-specified short name in the source code without writing the URL. Furthermore, to update the version of a module, you can change the version at once by rewriting the import maps. However, you can't specify the version by semver like NodeJs package.json, and you can't resolve dependency-hell.
> ### CDNs for NPM packages
> 
> Skypack, Esm.sh, and Esm.run are CDN services that convert packages uploaded on npmjs.com to ES Module and deliver them. These services compile transparently on the edge network. In addition, they compile and interpret the version directive specified in the URL and automatically deliver the most appropriate version of the package. However, We are not sure if these services will work. However, these services only deliver packages uploaded to npmjs.com, not necessarily projects registered on deno.land.
> ### Deno Package Managers
> 
> Package managers made by deno, such as deno-udd, fall into two main categories. One is to prepare an external file, like import-maps, and install modules based on the specified version. The other is to add `#` to the URL and manage the version based on the directive at the end. This method is not intended to solve dependency-hell and requires a separate software installation.
> ## Version Aliasing for Deno.land
> ### Semver
> 
> Semver is a simple set of rules and requirements that dictate how version numbers are assigned and incremented. It is commonly used in the JavaScript community. Although we have no statistical information on how many libraries follow this specification in Deno community, we believe this is a major manner because the Deno community is close to the JavaScript community. We propose the redirection of requests based on this rule as follows.
> 
>     1. The client sends HTTP request to the server to retrieve modules.
> 
>     2. The server searches git tags for the closest version of the library.
> 
>     3. The server rewrites the URL with the version and return the 302 redirections.
> 
>     4. The client sends HTTP request to the redirected URL.
> 
> 
> Our proposal is to implement step 2 and 3 based on deno-semver, which is compatible with package.json.
> ### Non-semver
> 
> We also concern that some of the libraries do not follow the semver spec. Hence, the system should fall back to be compatible with the existing system. In above step 2, if the program could not found the version because the request or the library did not follow the semver, the server does not rewrite a version number. Thus, the client sends the same request to deno.land.


kt3k:
hmm `dotland-fresh` preview is available, but `dotland` preview doesn't seem available.. 🤔

tani:
It's weird. For #2078, dotland preview is available but dotland-fresh preview is not.
On the other hand, dotland-fresh preview is available for #2077  but dotland preview is not.

kt3k:
pushed merge commit to re-trigger the preview deployment (but looks failed..

kt3k:
This PR optionally adds semver resolution to the registry if the url has `?semver` query string.

For example, by applying this PR:
- `https://deno.land/x/oak@^10.0.0/mod.ts?semver` resolves to `https://deno.land/x/oak@v10.4.0/mod.ts`
- `https://deno.land/x/lume@~1.6.0/mod.ts?semver` resolves to `https://deno.land/x/lume@v1.6.4/mod.ts`

I'm in favor of landing this PR and getting the feedback from the community.

Note: If the url doesn't have `?semver`, this doesn't affect anything. So this shouldn't have any effect to existing code in the ecosystem.

kt3k:
Closing this for now. Similarly to #1826, we currently don't have a plan to add feature like this in the registry. Thank you for your efforts @tani anyway.

PR #2070, opened 2022-03-17
feat(worker): use different GA endpoints for bot/deno vs human traffic

piscisaureus:


piscisaureus:
I have added `GA_TRACKING_ID_FOR_BOTS` and GA_TRACKING_ID_FOR_HUMANS` to the project's list of env vars.
Bot traffic goes to the "deno.land" property - we should rename it to "bots" after this PR is merged.
Human traffic goes to the "deno.com" property - we should rename it to "humans".

lambtron (approved):

PR #2075, opened 2022-03-19
Update Deploy button styles

jordanreger:
changed the deploy button styles so that it's not distracting and doesn't appear as if it's selected.


ry (approved):
Lgtm - thanks

PR #2077, opened 2022-03-20
hero gradient background

ry:
Could anyone help me use https://gist.github.com/ry/a0532ce538e6331947e6755e8f512b43 to do the background in SVG instead of JPEG?

Screen Shot 2022-03-20 at 5 28 14 PM


pages/index.tsx
```diff
@@ -23,7 +23,10 @@ serve((req) => new Response("Hello World\\n"), { port: 8000 });`;
         <title>Deno - A modern runtime for JavaScript and TypeScript</title>
       </Head>
       <div class="bg-white">
-        <div class="bg-gray-50 border-b border-gray-200">
+        <div
+          class="bg-gray-50 border-b border-gray-200"
+          style="background-image: url('/background.jpeg');"
+        >
```
meszarosdezso, line 29:
```suggestion
        <div
          class="bg-gray-50 border-b border-gray-200 relative"
        >
          <Background />
```

the `Background` component should have `position: absolute` and `inset: 0` to fill its closest relative parent (thats why the relative class is necessary on the div above)

meszarosdezso (commented):
Hey! From a quick glance I believe this should do the trick. I can give it a shot in the next few days if you want.

Also, how about making the background gradient fixed when scrolling, is that too cliche nowadays? 😅

PR #2092, opened 2022-03-28
add toolchain to home page

lambtron:
after talking to some deno users, some of them are surprised to learn about how robust and useful the built-in toolchain is. 

this is just a quick pass at featuring the toolchain in a more prominent way.

also, right now some links point to version 1.20.3 of the manual. is there a way to future proof linking to that particular version (e.g. is there a `version` variable that i can put into the anchor tag)?

### DIFF
--- a/pages/index.tsx
+++ b/pages/index.tsx
@@ -239,10 +239,10 @@ Emit "colors.bundle.js" (9.83KB)`;
           </a>
           <p class="my-4 text-gray-700">
             Deno comes with a robust{" "}
-            <a class="link" href="/manual/tools">set of tools</a>, so
-            you can spend less time searching and evaluating third party
-            modules, and more time writing code and being productive. Here are a
-            few examples.
+            <a class="link" href="/manual/tools">set of tools</a>, so you can
+            spend less time searching and evaluating third party modules, and
+            more time writing code and being productive. Here are a few
+            examples.
           </p>
           <p class="my-4 text-gray-700">
             <a class="link" href="/manual/tools/linter">Lint</a>{" "}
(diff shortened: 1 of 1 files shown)


crowlKats:
you can just not specify the version; that will then simply redirect to the latest available version

pages/index.tsx
```diff
@@ -207,6 +231,54 @@ serve((req) => new Response("Hello World\\n"), { port: 8000 });`;
...
+            more time writing code and being productive. Here are a few
+            examples.
+          </p>
+          <p class="my-4 text-gray-700">
+            <a class="link" href="/manual/tools/linter">Lint</a>{" "}
+            all JS/TS files in the current directory and subdirectories:
+          </p>
+          <p>
+            <CodeBlock
+              code={"deno lint\nChecked `n` files"}
```
bartlomieju, line 253: Replace with some actual value?

pages/index.tsx
```diff
@@ -207,6 +231,54 @@ serve((req) => new Response("Hello World\\n"), { port: 8000 });`;
...
+              language="bash"
+            />
+          </p>
+          <p class="my-4 text-gray-700">
+            <a class="link" href="/manual/tools/formatter">Format</a>{" "}
```
bartlomieju, line 258: Ditto

pages/index.tsx
```diff
@@ -207,6 +231,54 @@ serve((req) => new Response("Hello World\\n"), { port: 8000 });`;
...
+          <p class="my-4 text-gray-700">
+            <a class="link" href="/manual/tools/bundler">Bundle</a>{" "}
+            your project into a single JS file, including all dependencies:
+          </p>
+          <p>
+            <CodeBlock
+              code={denoBundleExample}
+              language="bash"
+            />
+          </p>
```
bartlomieju, line 276: I'd say bundler is quite controversial, since there's a mismatch in user expectation of what it can do. I suggest to replace it with `deno test` example

bartlomieju (approved):
LGTM

PR #2095, opened 2022-03-29
add wanted modules to x search page

lambtron:
- add call out to [wanted_modules](https://github.com/denoland/wanted_modules) in FAQ on /x
- add call out if no modules are found

image


ry (approved):
LGTM

crowlKats:
/x/ search is currently utterly bonkers and broken, it is likely we will get a decent amount requests for modules that actually already exist but cant be easily found

PR #2103, opened 2022-04-04
fix: use root readme if present

### LINKED ISSUE #2069: wrong readme selected to be shown
Expectation: When present, the root readme file is rendered in favour of other subpath-present readme files.

Reality: The first `README.md` file found in the dir listing regardless of it being in a subpath is used.

See https://deno.land/x/doauth where `docs/README.md` file is used, instead of the `README.md` in the root of the project.

panva:
This PR uses the relative project path length to sort the possible root readme candidates so that if present, the root one is used in favour of others.

resolves #2069

### DIFF
--- a/util/registry_utils.test.ts
+++ b/util/registry_utils.test.ts
@@ -198,6 +198,26 @@ Deno.test("findRootReadme", () => {
   }
 });
 
+Deno.test("findRootReadme selection", () => {
+  {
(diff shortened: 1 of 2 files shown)


panva:
It worked.

See https://dotland-fresh-d5s9pa6dt5j0.deno.dev/x/doauth vs https://deno.land/x/doauth

PR #2105, opened 2022-04-07
hero gradient background (svg)

meszarosdezso:
The same as #2077, but I recreated the background in SVG based on the original gradient in illustrator.

( I am not sure about the formatting changes, I ran `deno fmt` before commiting )

### DIFF
--- a/pages/index.tsx
+++ b/pages/index.tsx
@@ -32,7 +32,10 @@ test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out (27ms
         <title>Deno - A modern runtime for JavaScript and TypeScript</title>
       </Head>
       <div class="bg-white">
-        <div class="bg-gray-50 border-b border-gray-200">
+        <div
+          style="background-image: url('/hero.svg')"
+          class="bg-gray-50 border-b border-gray-200 relative bg-cover bg-center"
+        >
           <Header main />
           <div class="max-w-screen-sm mx-auto px-4 sm:px-6 md:px-8 pt-12 pb-20 flex flex-col items-center">
             <h1 class="font-extrabold text-5xl leading-10 tracking-tight text-gray-900">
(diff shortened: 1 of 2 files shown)


roj1512:
Why does it take a second for the gradient to appear? It is kinda annoying.

meszarosdezso:
@roj1512 you're right, it has to be a component so it loads with the rest of the content. thanks for noting, I updated the PR.

roj1512 (approved):

kt3k:
This seems breaking the menu in mobile screen

Before
スクリーンショット 2022-04-08 14 12 53

After
スクリーンショット 2022-04-08 14 11 44

pages/index.tsx
```diff
@@ -32,9 +33,10 @@ test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out (27ms
         <title>Deno - A modern runtime for JavaScript and TypeScript</title>
       </Head>
       <div class="bg-white">
-        <div class="bg-gray-50 border-b border-gray-200">
+        <div class="bg-gray-50 overflow-x-hidden border-b border-gray-200 relative">
+          <Background />
           <Header main />
-          <div class="max-w-screen-sm mx-auto px-4 sm:px-6 md:px-8 pt-12 pb-20 flex flex-col items-center">
+          <div class="z-10 relative max-w-screen-sm mx-auto px-4 sm:px-6 md:px-8 pt-12 pb-20 flex flex-col items-center">
```
roj1512, line 39:
```suggestion
          <div class="z-9 relative max-w-screen-sm mx-auto px-4 sm:px-6 md:px-8 pt-12 pb-20 flex flex-col items-center">
```

roj1512 (changes requested):

pages/index.tsx
meszarosdezso: Thanks, but unfortunately Tailwind [only has z-index classes](https://tailwindcss.com/docs/z-index) for each 10th value /:

Because it's relative, the z-index wasn't actually needed.

meszarosdezso:
Sorry for that @kt3k, fixed now.

kt3k (approved):
LGTM

pages/index.tsx
roj1512: That z-9 worked for me tho. Anyway, awesome to hear you fixed it!

meszarosdezso: yep, it worked because that class doesn't exists, so it has the same effect as removing it (: thanks for looking into it tho!

ry:
Let's try it!

Issue #2104, opened 2022-04-05
Pagination and Sorting


Industrial:
Hello.

Please add pagination and sorting to https://deno.land/x. Without it, you are looking through a list looking for the entries with the most stars, sorting them by stars in your head and then you can't go to the next page. This feels like a let down :-1:

Industrial:
Is there an API call that can be made to list packages that returns JSON?

lucacasonato:
The page has pagination: https://deno.land/x?page=3

lucacasonato:
> Is there an API call that can be made to list packages that returns JSON?

https://github.com/denoland/deno_registry2/blob/main/API.md#get-modules

meszarosdezso:
@Industrial I assume you are talking about _searching_ in the registry. Like on https://deno.land/x?query=deno there's no pagination or sorting indeed, but I agree it would be a nice UX improvement. I will look into the pagination part, I guess the sorting should be handled in [denoland/deno_registry2](https://github.com/denoland/deno_registry2) (?)

Industrial:
Yes, I am talking about sorting and pagination in the UI. Having sorting would make it easier to glance what the community is leaning towards for solving problems and pagination for knowing what else is out there.

crowlKats:
Search has been fixed

PR #2109, opened 2022-04-11
refactor(registry): extract pagination into a component

meszarosdezso:
This extracts the pagination on the registry page into a component and adds some types.

Might be useful for #2104


kt3k (approved):
LGTM, thanks!

PR #2110, opened 2022-04-11
fix: manual redirect handling in proxy

lucacasonato:
Fixes the website not redirecting deno.land/std to deno.land/std@0.134.0


kt3k (approved):
LGTM

kt3k:
Seems like we need to consume the result body to prevent sanitizer error in testing

PR #2111, opened 2022-04-14
feat(registry): add next page button for module search

meszarosdezso:
This adds a **"Next"** button to the bottom of the search results in the registry, so users can navigate to the next page of results without updating the url manually.

Screenshot 2022-04-14 at 14 22 30

As far as I could figure it out from the registry's source code, searching is sorting all the modules based on a scoring function, and returns the first 20, meaning we don't have a "number of results"  here and users can eventually get to a page with no related results.

Also, I've seen the api handler checks sorting by _newest_ / _oldest_ / _stars_ by url parameter, but that doesn't seem like working currently (https://deno.land/x?sort=newest).

towards #2104

### DIFF
--- a/components/Pagination.tsx
+++ b/components/Pagination.tsx
@@ -15,6 +15,14 @@ interface Props extends PaginationProps {
   };
 }
 
+function getNavButtonStyles(disabled: boolean) {
+  return `relative inline-flex items-center px-4 py-2 border border-gray-300 text-sm leading-5 font-medium rounded-md bg-white ${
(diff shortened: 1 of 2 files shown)


meszarosdezso:
The diff looks like a mess in Github's UI, It's actually just a check if the query existst and conditionally render the whole pagination or the next button. Also added a helper function for the button's tw classes.

kt3k (approved):
LGTM, nice! @meszarosdezso

PR #2113, opened 2022-04-16
feat(worker): Google Analytics 4 experiment

piscisaureus:


ry:
Can you move the ga4 stuff into deno.land/x/ga4 so we can share it between multiple projects?

How will this effect the current "deno for bots" GA account?

piscisaureus:
> Can you move the ga4 stuff into deno.land/x/ga4 so we can share it between multiple projects?

I will eventually - but there are still a number of kinks to iron out and I also this abstraction isn't right yet.
You can see the status quo here: https://analytics.google.com/analytics/web/#/p311458244/realtime/overview

> How will this effect the current "deno for bots" GA account?

For now it reports to both. The current GA account is unaffected.

worker/handler.ts
```diff
@@ -78,6 +80,111 @@ const ga: Reporter = async (...args) => {
...
+      return;
+    }
+
+    // Filter out traffic generated by uptime checks (Checkly and GCP).
+    if (
+      botName === "Checkly" ||
+      botName === "GoogleStackdriverMonitoring-UptimeChecks"
+    ) {
+      return;
+    }
```
ry, line 112: Maybe these should be filtered out elsewhere (i.e. during analysis)

If we rely on this, then a new check is added, or the UA changes or checkly, then the data will be thrown off and we won't really understand why.

worker/handler.ts
```diff
@@ -78,6 +80,111 @@ const ga: Reporter = async (...args) => {
...
+
+    const bot = isBot(request);
+    const botName = bot ? getBotName(request) : null;
+
+    // If the visitor is using a web browser, only create events when we serve
+    // a top level documents or download; skip assets like css, images, fonts.
+    if (!bot && !isNonAssetRequest(request, response) && error == null) {
+      return;
+    }
```
ry, line 104: Here too. It seems less error prone if we are non relying on these hardcoded heuristics.

completions_v2.ts
```diff
@@ -226,3 +226,26 @@ export const routes = {
...
+      }
+    }
+  }
+  return Promise.resolve(
+    new Response(null, {
+      status: 404,
+      statusText: "NotFound",
+    }),
+  );
+}
```
crowlKats, line 251: this is not necessary anymore

PR #2115, opened 2022-04-21
improve http server example

ry:
https://github.com/denoland/deno_std/pull/1641 should be landed so this boilerplate can be removed.

Aiming for ultimate simplicity:
```js
import { serve } from "https://deno.land/std/http/server.ts";
serve(req => new Response("Hello World\n"));
```

### DIFF
--- a/pages/index.tsx
+++ b/pages/index.tsx
@@ -11,9 +11,7 @@ import versions from "../versions.json" assert { type: "json" };
 import { Background } from "../components/HeroBackground.tsx";
 
 export default function Home() {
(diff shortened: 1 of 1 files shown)


vwkd:
Not convinced that this “ultimate simplicity” of deleting 1 LOC is worth the added complexity to get rid of the baked in log message, like in customer applications that shouldn’t log to console, serverless environments like Deno Deploy, etc. Longer argument [here](https://github.com/denoland/deno_std/pull/1641#issuecomment-989042783)

dsherret (approved):
LGTM, but I think it would be good to have a blank line between the import declaration and the statement.

Issue #2124, opened 2022-04-29
og:image conflicts with dark mode


mashaal:
The default image for sharing pages on other platforms conflicts with dark mode, and is not in a similar landscape aspect-ratio that many platforms expect.

Dark mode slack:
<img width="599" alt="image" src="https://user-images.githubusercontent.com/6100364/165888969-3e1635c8-3ac3-4e10-82c1-b10b40a7e588.png">

Ideally this image would not have an alpha channel, and would be a landscape ratio?

PR #2126, opened 2022-04-29
update og-image to an actual og-image

lambtron:
swapped out the current og image (a square icon) to a simple og image with proper dimensions.


kt3k (approved):
LGTM, nice!

kt3k:
Confirmed facebook renders this nicely!

スクリーンショット 2022-04-30 21 14 12

But twitter preview doesn't seem using the new image as far as I checked it in the [card validator](https://cards-dev.twitter.com/validator)

スクリーンショット 2022-04-30 21 16 13

We might need further meta tags for better rendering in twitter? 🤔

mashaal:
will close #2124

PR #2127, opened 2022-04-30
feat: release countdown

meszarosdezso:
This adds a release countdown to the landing page.

After the vote on discord and some iterations I think this is the most simple and responsive version, and works well with the rest of page. Open to feedback, let me know what you think!

Desktop: 

Screenshot 2022-04-30 at 10 49 59

Mobile:

Screenshot 2022-04-30 at 10 50 27

For keeping it up-to-date I added a field to `versions.json` to update the date on every release. If this workflow is ok for the maintainers I can add some info about it to `cut_a_release.md` in the main repo. 

An other idea I had is to scrape the date from the manual page, which sounds a bit hacky and probably would slow down the initial render, but wouldn't need to update it *manual*ly.

Im open to better ideas.


versions.json
```diff
@@ -302,5 +302,6 @@
     "0.36.0": "0.36.0",
     "0.35.0": "0.35.0",
     "0.34.0": "0.34.0"
-  }
+  },
+  "next": "2022-05-18"
```
ry, line 306: @meszarosdezso This is nicely implemented but I don't want to add more complications to [our release process](https://github.com/denoland/deno/blob/main/tools/cut_a_release.md). Maintaining this line is extra work for us. 

Thank you for the feature, but closing without merge.

PR #2131, opened 2022-05-04
Add Deno Shield to showcase

ydcjeff:
Add https://shield.deno.dev to showcase


ydcjeff:
> nice site! heads up that it wasn't obvious to me to hit "enter" to generate the shield before i can click "copy markdown".

Thanks for the feedback. will improve it.

PR #2139, opened 2022-05-06
feat: readme switch view between rendered and code

crowlKats:
Similar to github's buttons to switch the view kind on markdown files. This UI can then also be re-used for when we want to switch between doc mode

https://user-images.githubusercontent.com/13135287/167098303-6aa1a0d0-d834-4d99-84c2-47be13bf7e7e.mov

### DIFF
--- a/components/FileDisplay.tsx
+++ b/components/FileDisplay.tsx
@@ -24,7 +24,7 @@ export function FileDisplay(props: {
 }) {
   const filetype = props.filetypeOverride ?? fileTypeFromURL(props.sourceURL);
   const filename = fileNameFromURL(props.sourceURL);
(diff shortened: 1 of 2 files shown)


components/FileDisplay.tsx
```diff
@@ -23,10 +24,13 @@ export function FileDisplay(props: {
 }) {
   const filetype = props.filetypeOverride ?? fileTypeFromURL(props.sourceURL);
   const filename = fileNameFromURL(props.sourceURL);
-
+  console.log(props.showCode);
```
kt3k, line 27: nit: remove?

kt3k (approved):
LGTM with nits

PR #2146, opened 2022-05-10
Add hindi language translation

elparthka:


elparthka (commented):
Add hindi language translation

kt3k (approved):

PR #2149, opened 2022-05-10
refactor: use tw function

crowlKats:
For some reason the menu bar on main page changed a bit


kt3k:
What's the benefit of this change?

crowlKats:
Its necessary for sharing of components between dotland and docland

kt3k (approved):
> Its necessary for sharing of components between dotland and docland

Sounds reasonable

LGTM

PR #2155, opened 2022-05-16
refactor: use islands

crowlKats:


deps.ts
```diff
@@ -1,34 +1,51 @@
...
+    extend: {
+      lineHeight: {
+        0: "0",
+      },
+      /*fontFamily: {
+        sans: ["Inter var", ...defaultTheme.fontFamily.sans],
+      },*/
+      width: {
+        72: "18rem",
+      },
```
ry, line 37: What is 72 here?

islands/VersionSelect.tsx
```diff
@@ -0,0 +1,33 @@
...
+            location.href = versions[e.currentTarget.value];
+          }
+        }}
+      >
+        {!Object.hasOwn(versions, selectedVersion) && (
+          <option key={selectedVersion} value={selectedVersion}>
+            {selectedVersion}
+          </option>
+        )}
+        {Object.keys(versions).map((tag) => (
```
ry, line 25: Objects don't have an ordering  - presumably we want the versions ordered. Maybe use `[string,string][]`instead?

lucacasonato: > Objects don't have an ordering

They do. Since ES2015 ordering of string keys in objects follow insertion order.

lucacasonato:
Please set `disabled=true` on the SSR render of the dropdown.

ry (approved):
LGTM

PR #2157, opened 2022-05-17
feat(showcase): lazy load images

ydcjeff:
Although it might not improve time to first byte very much, it is still good to have.


kt3k (approved):
LGTM. Thanks!

PR #2158, opened 2022-05-17
fix: add missing labels to y axes

Rauno56:
On [benchmarks](https://deno.land/benchmarks), adding a label to y axes for:

* Thread count
* Syscall count and
* Cargo Dependencies

Currently displays "undefined".


Rauno56:
I can close this if this is the expected behavior:
image

kt3k (approved):
Sorry for the delay in review! Good catch!

LGTM

PR #2177, opened 2022-06-12
Improve manual routing

### LINKED ISSUE #2164: Manual/Introduction link inconsistencies lead to 404 page
When on the Manual's ["Introduction"](https://deno.land/manual) page -- coming from the home page's topbar link -- clicking on any of the links in the [Feature highlights](https://deno.land/manual#feature-highlights) section redirects users to a 404 page.

For example, clicking on [web platform functionality](https://deno.land/runtime/web_platform_apis.md) redirects to [https://deno.land/runtime/web_platform_apis.md](https://deno.land/runtime/web_platform_apis.md) (a 404), rather than [https://deno.land/manual/runtime/web_platform_apis](https://deno.land/manual/runtime/web_platform_apis) (the correct one).

**Notably**, the issue occurs only when starting from [https://deno.land/manual](https://deno.land/manual) but works fine when starting from [https://deno.land/manual/introduction](https://deno.land/manual/introduction). For further information, the first link is what we get when you click "Manual" from the [home page](https://deno.land)'s links, and the second is what we get if we click "Introduction" in the table of contents sidebar when already on the Manual page.

I've tested on Safari, Chrome, and Firefox.

Seems to be similar to #2128, but the links specified there seem to have been fixed. Also, I think the main issue here is the inconsistency between [https://deno.land/manual](https://deno.land/manual) and [https://deno.land/manual/introduction](https://deno.land/manual/introduction).

kidonng:
- Redirect `/manual` to `/manual/introduction`, fix #2164
- Strip `.md` from manual content to avoid extra redirect requests. This is safe since I checked the manual and didn't find any `.md` that isn't a internal link.


kidonng:
Also, I want to propose the idea to always version the manual URL as well (i.e. `https://deno.land/manual` always redirect to a specific version `https://deno.land/manual@va.b.c`). The same manual link may show different content as new versions are released, which is not consistent with how the module pages works.

lukemcdo:
Without the redirection, the https://deno.land/manual page has numerous broken links. Is this open because the issue is in progress, or should I open a new issue?

routes/manual.tsx
```diff
@@ -356,8 +356,9 @@ export default function Manual({ params, url, data }: PageProps<Data>) {
               <div class={tw`pt-1`}>
                 <Markdown
                   source={data.content
-                    .replace(/\$STD_VERSION/g, stdVersion)
-                    .replace(/\$CLI_VERSION/g, version)}
+                    .replaceAll(".md", "")
```
crowlKats, line 359: this doesnt seem ideal; what if we mention .md files in some manual entry? I think we need to figure out why the URLs have .md in the first place, and remove that

kidonng: > what if we mention .md files in some manual entry?

We can make the regex only match markdown links (`[text](link)`) and that should avoid most false positives.

> I think we need to figure out why the URLs have .md in the first place, and remove that

If the file extension is removed, the link becomes invalid in the _source_ file. I think it's better to do transformation here.

kidonng:
Hmm, [CI encountered 500 on GitHub raw](https://github.com/denoland/dotland/runs/7533616765?check_suite_focus=true#step:6:659)... I think we really should stop importing stuff from GitHub raw and use a CDN (like jsDelivr) instead.

crowlKats:
@kidonng can you rebase? also, could you add a condition that if no version is specified, redirect to latest version?

crowlKats:
going to `/manual` redirects to `/manual@v1.24.3/introduction`, but going to `/manual@v1.24.3` doesnt

routes/manual.tsx
```diff
@@ -277,15 +272,21 @@ function ToC({
...
       url.pathname = url.pathname.slice(0, -3);
       return Response.redirect(url);
     }
 
-    const version = params.version || versions[0];
     const sourceURL = getFileURL(
       version,
-      params.path ? `/${params.path}` : "/introduction",
+      `/${params.path}`,
     );
```
crowlKats, line 290: nitpick:
```
    const sourceURL = getFileURL(version, `/${path}`);
```
(github doesnt allow me to comment as suggestion)

crowlKats (approved):
LGTM

routes/manual.tsx
kidonng: I guess `deno fmt` is less smart than Prettier 🥲

PR #2182, opened 2022-06-16
ci: type check only local files

kt3k:
Main CI is failing with remote type errors, which we can't fix on our side. https://github.com/denoland/dotland/runs/6913536201?check_suite_focus=true

I think local type checking is enough for the CI for this repository. What do you think @crowlKats ?


satyarohith (approved):
LGTM. I agree that we should check only local files as remote are out of our control.

crowlKats (approved):
Makes sense

PR #2185, opened 2022-06-16
Fix icon size and margin

kidonng:
Bug introduced in #2179:

image

- The sizes of the three icons are inconsistent
- Star icon has wrong margin

[What it used to look like](https://dotland-31pv120xdm80.deno.dev/std@0.144.0):

image


bartlomieju (approved):
Thanks

PR #2202, opened 2022-06-24
use break-word instead of break-all

mooxl:
I noticed that the inline code can jump to a new line after each letter.
Perhaps you have your reasons for this, but in my opinion this is not so nice to look at and hard to read.

Index:
Screenshot 2022-06-24 at 11 12 29

Benchmarks:
Screenshot 2022-06-24 at 11 17 09


crowlKats (approved):
thanks!

PR #2204, opened 2022-06-24
fix: center sidebar

### LINKED ISSUE #2189: Center the current file in the sidebar by default
I loaded up https://deno.land/manual/examples/read_write_files via search and had no frame of reference for where I was in the docs.

The request would make: 

image

Instead be:

image

mooxl:
Resolves #2189


crowlKats (approved):
LGTM, thanks

PR #2213, opened 2022-06-27
chore: remove unnecessary `catch`

roj1512:
This is because `report.send` won't throw errors.[1]

[1] https://github.com/denoland/ga4/blob/04a1ce209116f158b5ef1658b957bdb109db68ed/mod.ts#L227


crowlKats (approved):
Thanks

PR #2224, opened 2022-06-30
chore: make docs only appear on std

crowlKats:


lucacasonato (commented):
Data fetching for the doc view still happens for /x. The whole point is to avoid that.

crowlKats:
Done...

bartlomieju (approved):
LGTM, please revert if there are any troubles

PR #2241, opened 2022-07-06
fix: add code view button

crowlKats:


hyp3rflow:
Thanks for your fast feedback :)

components/DocView.tsx
```diff
@@ -85,3 +90,30 @@ export function DocView({
...
+  sourceURL: string;
+  children: DocNode[];
+  url: string;
+}) {
+  return (
+    <div>
+      <div class={tw`flex justify-between mb-8`}>
+        <div>{/* TODO: add module name */}</div>
+        <a
+          href={sourceURL + "?code"}
```
hyp3rflow, line 108: @crowlKats It seems that the codeview is not working correctly when it is in the main module.
For example, [std@0.147.0/http](https://dotland-xk6s5rf6hb1g.deno.dev/std@0.147.0/http) → [std@0.147.0/http?code](https://dotland-xk6s5rf6hb1g.deno.dev/std@0.147.0/http?code), not http/mod.ts?code

crowlKats: code view changes only the view, not file. doc view shows possible main module, whereas code view just shows the code as-is. unsure if we should change this, but as it is now is simplest.

hyp3rflow: Okay, it makes sense!

kitsonk, line 108:
```suggestion
          href={`${sourceURL}?code`}
```

kitsonk (approved):
LGTM, one minor nit

PR #2246, opened 2022-07-08
refactor: redesign module path index

crowlKats:


ije:
Screen Shot 2022-07-08 at 12 43 12
seems those link color is not correct

ije (approved):
Looks great 👍

PR #2254, opened 2022-07-10
feat: symbol doc view

crowlKats:


lucacasonato (commented):
I am not fond of this ~ special casing. It's going to break when someone creates a folder called ~ in a module.

This is also going to make the regexp match for one of the most common routes in the project much more expensive.

crowlKats:
> I am not fond of this ~ special casing. It's going to break when someone creates a folder called ~ in a module.

well, we use the same in docland. would you prefer a search param? not that fond of that tbh, though it would also solve your second point

> This is also going to make the regexp match for one of the most common routes in the project much more expensive.

yea, it is quite annoying that a lazy matcher isnt possible with urlpattern currently. maybe something worth investigating to add to spec.

kitsonk:
> I am not fond of this ~ special casing. It's going to break when someone creates a folder called ~ in a module.

What other sigil would you recommend?  I don't want us to use query parameters, as they need to be hard links to a page.  Realise they will always be `{path-to-module}/~/{symbol}`.

doc_components/styles.ts
```diff
@@ -97,12 +97,11 @@ const styles = {
     `block pl-2 py-0.5 lg:(table-cell w-48 pl-4 pr-2 py-1.5)`,
   moduleIndexPanelCell: apply`block lg:table-cell ${panelStyles}`,
   modulePathIndex: apply`rounded-lg w-full border border-secondary`,
-  modulePathIndexHeader: apply`flex justify-between items-center`,
+  modulePathIndexHeader: apply`flex justify-between items-center py-3.5 pr-5`,
+  SymbolDocHeader: apply`flex justify-between items-center py-3.5`,
```
kitsonk, line 101: We should be consistent in casing:

```suggestion
  symbolDocHeader: apply`flex justify-between items-center py-3.5`,
```

Also, they are all alphabetically ordered.

main.ts
```diff
@@ -20,27 +20,30 @@ import options from "./options.ts";
...
-    namespace: string | undefined,
-    symbol: string,
+    _current: string,
+    _namespace: string | undefined,
+    _symbol: string,
   ): string | undefined {
+    return undefined;
     // FIXME(bartlomieju): special casing for std here is not ideal
-    if (current.startsWith("/std")) {
+    /*if (current.startsWith("/std")) {
```
kitsonk, line 38: Do we really want to leave commented out code in here?

kitsonk (approved):
A couple of minor nits

main.ts
crowlKats: i guess it is not necessary

PR #2255, opened 2022-07-10
Chore: add std versions back to std imports on homepage

### LINKED ISSUE #2245: The getting started examples on the home page should include std version
For some reason @ry removed the explicit version in this commit: https://github.com/denoland/dotland/commit/e8659b4931530133cc43c2f18d763cb895a10abe. This is error prone, and it should be added back. The referenced commit also shows how to add it back. This should be done for all std references on the home page.

nrademacher:
Closes #2245 by adding the version (back) to the `std` imports on the homepage.

I also added another commit (e6f7d822) where I replaced the hard-coded `std` version in two places, since users might be confused by the version mismatch. If this is undesired or too out-of-scope, however, I could throw out the commit.

### DIFF
--- a/routes/index.tsx
+++ b/routes/index.tsx
@@ -20,8 +20,9 @@ interface Data {
 }
 
 export default function Home({ data }: PageProps<Data>) {
-  const complexExampleProgram =
-    `import { serve } from "https://deno.land/std/http/server.ts";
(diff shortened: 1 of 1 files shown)


crowlKats (approved):
LGTM, thanks

PR #2261, opened 2022-07-12
fix: replace or remove invalid tailwind classes

crowlKats:


ije:
Screen Shot 2022-07-12 at 17 19 24
still got warns from my dev env

lucacasonato (approved):

Issue #2263, opened 2022-07-12
non-docable files error out doc page


crowlKats:
Any non-docable file will error out the page, ie https://deno.land/std@0.136.0/Releases.md. these should redirect to their code alternative. 
https://github.com/denoland/dotland/blob/22925d0ee7f8e0b5c1669ebc9aaa8b7c7843e940/util/doc.ts#L36

kidonng:
I was going to report this a few days ago 😅 note this only applies to `/std`, not `/x`.

crowlKats:
yea, though this would apply to `/x` as soon as we enable docs there, which hopefully isnt too far off. quick-fix would be try-catch the whole content of the function and return null if there is an error

crowlKats:
Closing as addressed

PR #2268, opened 2022-07-13
Fix doc page error on non-docable files

kidonng:
Temporary fix regarding #2263, as suggested in https://github.com/denoland/dotland/issues/2263#issuecomment-1182685975.


crowlKats (approved):
LGTM, thanks!

PR #2270, opened 2022-07-13
docs: fix dead Web APIs link

mxdvl:
I think the `.md` extension is erroneous.

You could run some automated tasks to make sure all your URLs are valid.


crowlKats:
The manual handler takes care of md extension, so it isnt wrong, but either way this is nicer

crowlKats (approved):

mxdvl:
> The manual handler takes care of md extension, so it isnt wrong, but either way this is nicer

This was a dead link on the website yesterday, which is how I was inclined to raise this PR. Can't seem to reproduce today, but I did get an error page with the pixel dinosaur and crab.

PR #2280, opened 2022-07-19
fix: prevent twind preflight from overwriting img and video elements height attribute

### LINKED ISSUE #2279: Invalid image size rendering in README's
`deno.land/x` seems to overwrite height of images to `auto` which breaks images in README that specify `width` or `height` attributes.

On the top how it's supposed to look like (how github renders it) and on the bottom how it looks like.

image
image

Im-Beast:
Closes #2279

### DIFF
--- a/util/twind.ts
+++ b/util/twind.ts
@@ -5,12 +5,11 @@ import { css } from "twind/css";
 export { css };
 
 export const config: Configuration = {
-  preflight: {
-    body: apply`text-default ${
-      css({
-        "scroll-behavior": "smooth",
-      })
-    }`,
+  preflight(preflight) {
+    delete preflight["img,video"];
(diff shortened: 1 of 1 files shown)


crowlKats:
unsure if this really is the right approach; this should rather be fixed "properly"

Im-Beast:
This is "proper" approach or at least I don't see other way around it.

Preflight applies `max-width: 100%` and `height: auto;` to `img,video` and if either height or width is present in CSS size attributes don't really work.

crowlKats (approved):
thanks

crowlKats:
@Im-Beast could you rebase?

Im-Beast:
Will do right now

Im-Beast:
Done @crowlKats

PR #2289, opened 2022-07-22
fix(search): handle ctrl+k and / shortcuts

crowlKats:


lucacasonato (commented):
Please add a check to ensure that the user does not have an `input` field focused at the time of entering the `/`, because otherwise we will open the search while they are typing in some other field.

crowlKats:
@lucacasonato done

lucacasonato:
doesn't work after you select then unselect an input field

crowlKats:
it does for me..

lucacasonato (approved):
mh, now it's working for me too

PR #2296, opened 2022-07-23
refactor: use header in manual

crowlKats:
Manual redesign will come after this


kitsonk:
Because the headers have different max widths, there is a big jump when moving between them, which is quite jarring:

Introduction___Manual___Deno

std_0_149_0___Deno

crowlKats:
@kitsonk I made it full-size again.
Also just noticed a bug with mobile view, with get that fixed ASAP

kitsonk (commented):
LGTM pending mobile view issue fix

kitsonk (approved):

PR #2299, opened 2022-07-24
fix: add more image formats

### LINKED ISSUE #2298: `deno.land/x` doesn't display WebP and AVIF files
Example:

https://deno.land/x/goodgle@0.7.0/assets/instructions/brave/1.webp?code

https://deno.land/x/goodgle@0.7.0/assets/instructions/brave/1.avif?code

crowlKats:
Fixes #2298

### DIFF
--- a/util/registry_utils.ts
+++ b/util/registry_utils.ts
@@ -490,7 +490,7 @@ export function fileTypeFromURL(filename: string): string | undefined {
     return "markdown";
   } else if (f.match(`\\.${orgExtension}$`)) {
     return "org";
-  } else if (f.match(/\.(png|jpe?g|svg)/)) {
+  } else if (f.match(/\.(png|jpe?g|svg|webm|webp|avif)/)) {
     return "image";
   }
 }


kitsonk:
I'm not sure we really want deno.land/x to be an image browser. As discussed recently, displaying indexes of directors (file browsing) is deprecated behaviour. Rendering code that a stack trace might include so it can be linked to is one thing, but arbitrary media browsing doesn't seem like a good idea. What need in discovering and understanding the module does it fill?

 @lucacasonato @ry thoughts?

kidonng:
> displaying indexes of directors (file browsing) is deprecated behaviour

Even though modules in `deno.land/x` come from GitHub, I don't think the two are fully equivalent in file browsing. For example, consider that modules on `deno.land/x` are immutable, but Git tags can be retagged, the repository may vanish, etc.

But I don't really have a say in this, it is probably fine not to display media on `deno.land/x`, I just think this is wrong:

image

> Rendering code that a stack trace might include so it can be linked to is one thing

Regarding this, I just found it is currently broken: #2300

kt3k (approved):
LGTM

PR #2307, opened 2022-07-26
Fix markdown table overflow

kidonng:
https://github.com/denoland/dotland/pull/2304#issuecomment-1195477952

### DIFF
--- a/routes/gfm.css.ts
+++ b/routes/gfm.css.ts
@@ -11,6 +11,9 @@ const css = `${CSS}
 .markdown-body ol {
   list-style: numeric
 }
+.markdown-body table {
+  width: fit-content;
+}
 `;
 
 export const handler: Handlers = {


kidonng:
There are some markdown-related styles here:

https://github.com/denoland/dotland/blob/f583ec6b6d86b36e49f16a229223ee7ce3f7223f/static/app.css#L145-L155

Would you like to see them moved into `gfm.css.ts`?

crowlKats:
the thing is, this wasnt necessary in the previous design, and nothing with the rendering has changed, so there is def a solution without modifying app.css

crowlKats:
Closing in favour of #2310

kidonng:
> the thing is, this wasnt necessary in the previous design, and nothing with the rendering has changed, so there is def a solution without modifying app.css

I may be misleading, this pull request doesn't change `app.css` (it's `gfm.css.ts`), I just saw there are some (duplicate?) styles that may be worth cleaning up.

crowlKats:
regardless if it's app.css or gfm.css.ts, it wasnt necessary before. i will merge this PR as-is and look into it on a later date for a proper fix

crowlKats (approved):

PR #2329, opened 2022-07-28
chore: pin esm.sh version for dependencies

cjihrig:
The latest version appears to have broken something and is blocking the v1.24.1 release.


cjihrig:
Closing as it doesn't look like all esm.sh dependencies can be pinned.

PR #2334, opened 2022-07-30
refactor: redesign /x/ landing page

crowlKats:
the total modules counter is currently disabled; will enable it in a follow-up to switch to algolia search


ije:
let's use twind's default font family, currently the font will tingle from `serif` to `sans-serif` when load is slow

routes/x/index.tsx
```diff
@@ -42,391 +33,242 @@ export default function ThirdPartyRegistryList({ url, data }: PageProps<Data>) {
...
-                  <dd class={tw`mt-2`}>
-                    <p
-                      class={tw`text-base leading-6 text-gray-500 break-words`}
+          <div class={tw`border border-dark-border rounded-lg overflow-hidden`}>
+            <form
+              class={tw`px-5 py-4 flex items-center justify-between border-b border-dark-border bg-ultralight leading-none`}
+              method="get"
+            >
+              <span class={tw`font-semibold`}>Explore Modules</span>
+              <input
```
ije, line 99: miss "glass" icon, can we use the same search input component of the global search

crowlKats: oh ok, the figma design didnt have the magnifying glass, thats why i left it out

ije: my mistake

crowlKats:
@ije fixed. regarding the font, i added a fallback sans-serif font, so that should make the flash less evident. we kinda need to use this font right now because the arrow in the footer for job application relies on a ligature available in that font

kitsonk:
What would it take to change the list to something ordered by the popularity score?  While the design looks great, the impression of the list with lodash at the top is really a problem.

crowlKats:
@kitsonk I will switch to use the algolia-based search (so same algolia index as what the global search uses) in a follow-up, that would fix that problem

ije:
> because the arrow in the footer for job application relies on a ligature available in that font

why not use icon instead, "inter" font is also missed in windows for most users. or can we add inter web font?

crowlKats:
we could add an import for the font, but yea, maybe icon makes most sense?

ije:
i added the icon in figma

crowlKats:
@ije done

ije:
and there are other too UX things in "add module" page can be improved i think:

1. set the error text "Invalid Name/Name has been taken!" a fixed height to avoid the page tingle
2. let's move the "waiting..." spin to the bottom (or maybe only show after copy), please check the figma

crowlKats:
can I keep those changes for a follow-up PR?

routes/x/index.tsx
```diff
@@ -42,391 +33,251 @@ export default function ThirdPartyRegistryList({ url, data }: PageProps<Data>) {
...
+              <h1 class={tw`font-bold text-3xl text-black`}>
+                Deno Third Party Modules
+              </h1>
+
+              <p class={tw`leading-5 text-[#6C6E78]`}>
+                <span class={tw`font-semibold text-default`}>deno.land/x</span>
+                {" "}
                 is a hosting service for Deno scripts. It caches releases of
                 open source modules stored on GitHub and serves them at one easy
                 to remember domain.
```
roj1512, line 55:
```suggestion
                open source modules stored on GitHub and serves them at an
                easy-to-remember domain.
```

routes/x/index.tsx
```diff
@@ -42,391 +33,251 @@ export default function ThirdPartyRegistryList({ url, data }: PageProps<Data>) {
...
+            <div class={tw`space-y-4 w-full`}>
+              <h1 class={tw`font-bold text-3xl text-black`}>
+                Deno Third Party Modules
+              </h1>
+
+              <p class={tw`leading-5 text-[#6C6E78]`}>
+                <span class={tw`font-semibold text-default`}>deno.land/x</span>
+                {" "}
                 is a hosting service for Deno scripts. It caches releases of
                 open source modules stored on GitHub and serves them at one easy
```
roj1512, line 54:
```suggestion
                open-source modules stored on GitHub and serves them at one easy
```

roj1512:
@crowlKats Can't we add some filtering feature for the search results or just make the more-relevant results naturally on the top?

crowlKats:
as i have stated before, search will be fixed in a follow-up...

roj1512:
ah, alright.

ije (approved):
LGTM! nice job!

PR #2337, opened 2022-08-01
fix: better instatus badge

### LINKED ISSUE #2323: Unexpected black frame around the instatus button in the footer

kt3k:
Fixes the instatus badge in the footer (for the users who set OS default dark theme.)

Before
スクリーンショット 2022-08-01 15 01 24

After
スクリーンショット 2022-08-01 15 00 48

closes #2323

### DIFF
--- a/components/Footer.tsx
+++ b/components/Footer.tsx
@@ -71,11 +71,11 @@ export function Footer() {
           <div class={tw`space-y-5 w-60`}>
             <iframe
               src="https://denostatus.com/embed-status/light-sm"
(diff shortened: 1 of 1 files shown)


kt3k:
> hm its not perfect, you can still see bits of black on the corner; instatus really needs to fix this on their end, but for now i guess this will do

Right. I guess an ideal solution might be that they support something like `https://denostatus.com/embed-status/light-sm?color-scheme=light` which override the `color-scheme` of iframe html

PR #2339, opened 2022-08-01
feat: switch to page endpoint

crowlKats:
Depends on denoland/doc_components#19
Closes: #2341


routes/x/module.tsx
```diff
@@ -350,59 +297,90 @@ function VersionSelector({
...
+      return new Response(undefined, {
+        headers: {
+          Location: getModulePath(
+            name,
+            version,
+            newPath,
+          ),
+        },
+        status: 301,
+      });
```
crowlKats, line 360: one "nasty" aspect about this is that it also affects imports, so importing `https://deno.land/std@0.150.0/collections` will work as it redirects to `https://deno.land/std@0.150.0/collections/mod.ts`. i will fix this.

components/DocView.tsx
```diff
@@ -5,118 +5,94 @@
...
+              return (
+                <ModuleDoc
+                  url={url.href}
+                  sourceUrl={url.href}
+                >
+                  {data.docNodes}
+                </ModuleDoc>
+              );
+            case "file":
+              return <div></div>;
```
crowlKats, line 80: this needs to be handled somehow.

routes/x/module.tsx
crowlKats: should we respond with an error for raw as a directory is being requested?

kitsonk: for raws, it should be a 404 I think, or a 406 Not Acceptable (which means we can't negotiate content to provide a response, which is actually more accurate)

kitsonk:
The arrows are displaying in the nav, but aren't expandable, is that expected?

std_0_150_0___Deno

kitsonk:
Something is not right with code view of a dir: https://dotland-43gn4xkhvyag.deno-staging.dev/std@0.150.0/async?code returns all the files in the repo... I will check API land, but there is something wrong too with the display:

std_0_150_0___Deno

kitsonk:
The nav doesn't meet my user expectations at all... I don't expect to drop back to the root when collapsing a section, and I don't expect expanding another module to cause me to navigate to it:

https://user-images.githubusercontent.com/1282577/182728402-34fc516c-5134-420e-b330-1e485bf5e8c9.mp4

crowlKats:
Yes, the arrows arent openable, they only open when on a symbol for given file. there is no point in being able to open a list of symbols when you are on a page on which the main view is a list of symbols

kitsonk:
> Yes, the arrows arent openable, they only open when on a symbol for given file. there is no point in being able to open a list of symbols when you are on a page on which the main view is a list of symbols

But what about the other modules on that page?

kitsonk:
https://dotland-43gn4xkhvyag.deno-staging.dev/std@0.150.0/collections/mod.ts?s=BinaryHeap and https://dotland-43gn4xkhvyag.deno-staging.dev/std@0.150.0/collections/mod.ts?s=BSTree are returning errors and not sure why

crowlKats:
> Something is not right with code view of a dir: https://dotland-43gn4xkhvyag.deno-staging.dev/std@0.150.0/async?code returns all the files in the repo... I will check API land, but there is something wrong too with the display:

seems that `CodePageDirEntry.size` can be undefined

crowlKats:
> https://dotland-43gn4xkhvyag.deno-staging.dev/std@0.150.0/collections/mod.ts?s=BinaryHeap and https://dotland-43gn4xkhvyag.deno-staging.dev/std@0.150.0/collections/mod.ts?s=BSTree are returning errors and not sure why

```
TypeError: properties is not iterable
    at getClassItems (https://raw.githubusercontent.com/denoland/doc_components/5455c13b7e6176324357e58cf3b554542ed412bb/classes.tsx:71:14)
    at Object.CodeBlockClass (https://raw.githubusercontent.com/denoland/doc_components/5455c13b7e6176324357e58cf3b554542ed412bb/classes.tsx:189:17)
    
   ```
deno_doc issue?

**Comment by crowlKats** [comment]
> and why provide a visual indication something is expandable, but isn't... if we don't want to list the symbols, then we should not have an arrow for that module

aight, would you prefer removing indicator or being able to open them?

**Comment by kitsonk** [comment]
> > and why provide a visual indication something is expandable, but isn't... if we don't want to list the symbols, then we should not have an arrow for that module
> 
> aight, would you prefer removing indicator or being able to open them?

@ije thoughts?  My feeling is remove them, (but not shift the label back, so empty space)

**Comment by kitsonk** [comment]
> [BinaryHeap](https://dotland-43gn4xkhvyag.deno-staging.dev/std@0.150.0/collections/mod.ts?s=BinaryHeap)

Appears to be decoding doc nodes out of the datastore, which is strange, because _some_ classes are working... need to investigate further.

**Comment by ije** [comment]
I also think the arrow is just  used to expand/collapse the symbols without switch page

**Comment by kitsonk** [comment]
> > https://dotland-43gn4xkhvyag.deno-staging.dev/std@0.150.0/collections/mod.ts?s=BinaryHeap and https://dotland-43gn4xkhvyag.deno-staging.dev/std@0.150.0/collections/mod.ts?s=BSTree are returning errors and not sure why
> 
> ```
> TypeError: properties is not iterable
>     at getClassItems (https://raw.githubusercontent.com/denoland/doc_components/5455c13b7e6176324357e58cf3b554542ed412bb/classes.tsx:71:14)
>     at Object.CodeBlockClass (https://raw.githubusercontent.com/denoland/doc_components/5455c13b7e6176324357e58cf3b554542ed412bb/classes.tsx:189:17)
>     
> ```
> 
> deno_doc issue?

Fixed in https://github.com/denoland/apiland/pull/38

**Comment by crowlKats** [comment]
@kitsonk sidepanel toggling is now implemented

**Comment by crowlKats** [comment]
https://dotland-zmbnxadarwe0.deno.dev/std@0.150.0/collections/aggregate_groups.ts?s=aggregateGroups
```
TypeError: Cannot read properties of undefined (reading 'length')
    at Object.TypeParams (https://raw.githubusercontent.com/denoland/doc_components/8d1ffb95737d773c8a1bb9419c37101950258842/types.tsx:686:15)
    ```
seems another similar issue

kitsonk:
I know we can't do it yet on the landing page, but any chance we can get the popularity tag on the pages here while we work on the landing page (and actually we should always suppress the tag for `std` in my opinion:

std_0_151_0___Deno

kitsonk:
> @kitsonk sidepanel toggling is now implemented

there are things I am still not sure about with it, but feels like we can iterate from here.

crowlKats:
> I know we can't do it yet on the landing page, but any chance we can get the popularity tag on the pages here while we work on the landing page (and actually we should always suppress the tag for `std` in my opinion:
> 
> std_0_151_0___Deno

ah yea, will do that

crowlKats:
> there are things I am still not sure about with it, but feels like we can iterate from here.

sure; though any example of something you are unsure about?

ry:
The folders here end with "/"
Screen Shot 2022-08-04 at 5 32 41 PM

ry:
I think mod.ts should be above the folders and the mod.ts should be expanded by default so you can see the index.

crowlKats:
> mod.ts should be expanded by default so you can see the index.

by default you land on the mod.ts file, which is a page of an index of all symbols available; having also the sidepanel open by default makes little sense imo

kitsonk:
> sure; though any example of something you are unsure about?

yes, some of what Ryan is saying, but also, the target just being arrow, and the larger target causing navigation to occur instead of expanding the folder... feels like there should be some visual indication that you are going to navigate, versus just ending up on a different page half the time.

I haven't checked it on mobile, but now it is making me curious.

crowlKats:
> I haven't checked it on mobile, but now it is making me curious.

sidepanel on mobile iirc is currently disabled

crowlKats:
Fixed the trailing slash, improved the toggle area for module toggle and added a hover state; now design wise its nearly identical to the manual's sidepanel

kitsonk:
So expanding the symbols is still a bit clunky... it isn't keyboard navigable and when hovering, it lacks the hand to indicate it is actionable:

https://user-images.githubusercontent.com/1282577/182980522-8fe7faa4-1e5c-41a6-929c-c23b69d3ce46.mp4

Also, we need to figure out how to eliminate folders that don't have docable children or subfolders from the nav (I think that is an apiland thing) because clicking on the `docs` folder in the nav brings you here: https://dotland-30mgnt9pwa30.deno.dev/x/oak@v10.6.0/docs

We might also want to render that though a bit better if someone hard links to it, to something like "no documentation found" and give them a link back to the root of the module?

lucacasonato (changes requested):
Code serving should not be reliant on apiland being up. If a /x module is requested and `isHTML === false`, the source code should be served directly with no additional requests being made.

crowlKats:
yes, that is the plan. was going to do that later

crowlKats:
@lucacasonato actually, thats wrong: it has to be reliant on apiland given if you import without version

lucacasonato:
It should pull the version from s3, like we do now.

crowlKats:
currently we actually use the CDN. regardless; quite annoying having to add another request and handling for versioning with that...

kitsonk:
For now, for raw, let's pull the latest version from S3 if we need it.  We need a very stable and fast way to do this. I don't want to bring down every client just because there is a hiccup in apiland.  We will address this "hot path" in the migration of registry2 to the new infrastructure, but for raw, we might not get much better than what we have now.

crowlKats:
@lucacasonato fixed it, now raw code serving is pretty much the first thing that is done

crowlKats:
should we get rid of the CDN usages and use S3 directly? we have always used S3 for raw code serving, and cdn for rendered code/readme view, which makes little sense

lucacasonato:
yes, please switch to using s3 directly.

components/DocView.tsx
kitsonk: Let's just redirect to code view.

kitsonk:
Really bad layout: https://dotland-4c36wt9tfhkg.deno-staging.dev/x/lambda@1.24.2/mod.ts?s=ALBEvent

Feels that that should really be addressed.

kitsonk (approved):
LGTM, pending updating to the tagged version of doc components.

There are a few things I just noticed but let's do them as follow up fixes and get this out of the way.

PR #2340, opened 2022-08-01
fix: initially disable search

crowlKats:


islands/GlobalSearch.tsx
```diff
@@ -172,6 +172,7 @@ export default function GlobalSearch() {
       <button
         class={tw`pl-4 w-80 bg-[#F3F3F3] flex-auto lg:flex-none rounded-md text-light hover:bg-light-border`}
```
crowlKats, line 173: unsure how we should style the button when disabled

lucacasonato (approved):

PR #2342, opened 2022-08-01
cleanup global search

crowlKats:


islands/GlobalSearch.tsx
```diff
@@ -376,14 +395,17 @@ function ManualResult({ hierarchy, url, content }: ManualSearchResult) {
     title.push(entry);
   }
   return (
-    <a href={url} class={tw`block`}>
+    <a href={url} class={tw`flex items-center gap-4`}>
```
ije, line 398: can we change the background color a bit darker when mouse _hover_?

crowlKats: @ije done, however the perfect hover background color for odd items is the normal background color for even items; do you think that might be a problem?

ije (approved):
LGTM

PR #2346, opened 2022-08-03
[Add Artwork] Saudi Deno

X7md:
image

Preview:
Saudi Deno


data/artwork.json
```diff
@@ -309,5 +309,19 @@
       "github": "Arc-blroth"
     },
     "license": "CC-BY-SA-4.0"
+  },
+  {
+    "date": "2022-08-03",
+    "image": "/images/artwork/snow_den.png",
```
kt3k, line 315: This seems pointing to a wrong file?

Looks like you forgot to commit the actual image file?

X7md: My bad

X7md: Done, add image in `./images/artwork/`

kt3k (approved):
LGTM

PR #2347, opened 2022-08-03
refactor: use algolia for module listing

crowlKats:


crowlKats:
The one thing missing is how to get total number of modules

ry:
Why does it have 

import { start } from "[https://deno.land/x/fresh@1.0.0/server.ts](https://deno.land/x/fresh@1.0.0/server.ts?code)";

crowlKats:
@ry what do you mean?

ry:
Screen Shot 2022-08-04 at 5 26 26 PM

crowlKats:
ah, that is completely unrelated to this PR, however it has been fixed earlier and is on main already

ry:
can you merge with main?

kitsonk:
@crowlKats I added `popularity_tag` as a "display facet" to the algolia search, which means I believe it should be returned on the results now, with the same values of `top_1_percent`, `top_5_percent`, `top_10_precent`, or `undefined` that we use elsewhere.

kitsonk:
This is blocked on that we didn't have a filter to only include 3rd party modules in the list.  That is in a re-structured search index, but we need to integrate that back into this PR.

kitsonk (approved):
LGTM... awesome, this makes a huge difference.

PR #2355, opened 2022-08-08
fix: manual keyboard navigation

### LINKED ISSUE #2311: Manual sidebar acessibility
#2304 broke keyboard accessibility of the manual sidebar.

For example clicking on a chapter heading does not automatically open the details view for that chapter. The details view is automatically opened when a user navigates to a page inside of the chapter, but not when they are looking at the entrypoint for the chapter.

crowlKats:
Closes #2311

### DIFF
--- a/components/Icons.tsx
+++ b/components/Icons.tsx
@@ -191,10 +191,11 @@ export function ArrowRight(props: { class?: string }) {
 }
 
 export function TriangleRight(
-  props: { class?: string; tabindex?: number; onKeyDown?: string },
+  props: { class?: string; tabindex?: number; onKeyDown?: string; "aria-label"?: string },
 ) {
   return (
     <svg
+      aria-label={props["aria-label"]}
       class={tw(props.class ?? "")}
       tabindex={props?.tabindex}
       // @ts-ignore onKeyDown does support strings
--- a/components/Icons.tsx
+++ b/components/Icons.tsx
@@ -190,10 +190,15 @@ export function ArrowRight(props: { class?: string }) {
   );
 }
 
-export function TriangleRight(props: { class?: string }) {
+export function TriangleRight(
+  props: { class?: string; tabindex?: number; onKeyDown?: string },
+) {
   return (
     <svg
       class={tw(props.class ?? "")}
+      tabindex={props?.tabindex}
+      // @ts-ignore onKeyDown does support strings
+      onKeyDown={props?.onKeyDown}
       xmlns="http://www.w3.org/2000/svg"
       width="10"
       height="10"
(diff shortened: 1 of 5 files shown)


lucacasonato:
Nice, this is working well. Please add the right aria labels to the triangles.

components/Icons.tsx
```diff
@@ -190,10 +190,15 @@ export function ArrowRight(props: { class?: string }) {
...
 
-export function TriangleRight(props: { class?: string }) {
+export function TriangleRight(
+  props: { class?: string; tabindex?: number; onKeyDown?: string },
+) {
   return (
     <svg
       class={tw(props.class ?? "")}
+      tabindex={props?.tabindex}
+      // @ts-ignore onKeyDown does support strings
```
piscisaureus, line 200: Is it necessary to use a string though?

piscisaureus (approved):

components/Icons.tsx
crowlKats: Yes, else if we'd use a normal callback, it would have to be an island

routes/manual.tsx
```diff
@@ -244,55 +244,57 @@ function ToC({
...
+                    active ? "link bg-ultralight" : "hover:text-gray-500"
+                  } font-semibold` + (active ? " toc-active" : "")}
+                >
+                  <Icons.TriangleRight
+                    aria-label={`open section ${entry.name}`}
+                    onKeyDown="if (event.code === 'Space' || event.code === 'Enter') { this.parentElement.click(); event.preventDefault(); }"
+                    tabindex={0}
+                    class={"cursor-pointer" +
+                      (entry.children ? "" : " invisible")}
+                  />
```
lucacasonato, line 264: This can go entirely if you make `<summary>` a `display: list-item` instead of a `display: flex`.

routes/manual.tsx
```diff
@@ -244,55 +244,57 @@ function ToC({
...
-              <label
-                htmlFor={slug}
-                class={tw`flex items-center gap-2 px-2.5 py-2 rounded-md block ${
-                  active ? "link bg-ultralight" : "hover:text-gray-500"
-                } font-semibold` + (active ? " toc-active" : "")}
+              <details
+                open={active}
+                class="rotate-svg"
+                // @ts-ignore onToggle does support strings
+                onToggle={`event.currentTarget.firstChild.firstChild.setAttribute("aria-label", (event.currentTarget.open ? "close" : "open") + " section ${entry.name}")`}
```
lucacasonato, line 251: not necessary if the triangle icon goes.

routes/manual.tsx
crowlKats: the built-in triangle is worse, but ok...

kt3k (approved):
LGTM

crowlKats:
removed details & summary as bug with chrome

PR #2365, opened 2022-08-10
Remove google analytics

AnInternetTroll:
Due to growing concerns of google breaking european laws, as well as 3 countries now making it illegal, Google analytics should be removed to protect both the users's data, as well as the site owner from lawsuits. 

- Itally's statement: https://www.gpdp.it/web/guest/home/docweb/-/docweb-display/docweb/9782874#english
- An article about Austria: https://iapp.org/news/a/the-austrian-google-analytics-decision-the-race-is-on/
- France's statement: https://www.cnil.fr/en/use-google-analytics-and-data-transfers-united-states-cnil-orders-website-manageroperator-comply
- Netherlands is working on it, not final yet: https://www.autoriteitpersoonsgegevens.nl/sites/default/files/atoms/files/handleiding_privacyvriendelijk_instellen_google_analytics_april_22.pdf


ry:
We don't have access to individual IP addresses or other personal data.

AnInternetTroll:
> We don't have access to individual IP addresses or other personal data.

The issue is not you, it's Google and by extension the USA. Due to USA's laws the government can request European user's data if it has been transferred in the US. As such various countries have ruled that Google Analytics is unlawful.

Officialstrike:
Why not use something like [Plausible](https://github.com/plausible/analytics)? It's fully open source, easy to use and privacy-conscious unlike GA(even if Server-side), as we all know GA is a great tool however google grossly exploits people's privacy and forces people to be tracked while using deno.land is not ok. There are great alternatives which would be fine and still contain a lot of data, if Plausible doesn't have the analytics you need you can always go with [Matomo](https://github.com/matomo-org/matomo). Just please anything **but** Google Analytics, for the sake of Deno lovers, privacy advocates and anybody else who doesn't want to have google peering into everything they do.

roj1512:
@Officialstrike I’m with you, but we should also remember that Deno is not a project like Debian or Wikimedia, it is a company.

BrunoBernardino:
While I'm not fond of how tersely and quickly this was closed, I can sympathize with the fact that the Deno company (and @ry specifically) is more concerned with and focused on improving Deno itself (and Deno Deploy).

The best we can do is show (here, as a comment, for example) how much we care about this, so they can hopefully reconsider.

I'd like to see GA removed too.

Also, this isn't really about the tracking Google is making here (those who are concerned are already blocking it, most likely), but the signal it sends as a company when you're actively using and perpetuating products and services by such companies that harm privacy so much.

It's also perfectly fine by me if that's not a battle Deno want's to be a part of, though. Their battle is already a tough one to fight as it is.

jhechtf:
we definitely should remove GA.

Sorikairox:
Definitely agree with removing GA

PR #2372, opened 2022-08-12
fix: use https for getVersionList

crowlKats:


ry:
Can you provide some more information? What is this fixing?

crowlKats:
because it currently uses http and not https, publishing modules currently doesnt work as the clientside errors with mixed content.

lucacasonato (commented):
I don't get it. Why is this needed?

lucacasonato:
Oh missed the above comment.

lucacasonato:
Can't you proxy the browser requests through the dotland backend, rather than talking to s3 directly?

crowlKats:
you mean adding a route to act as a proxy? it could be done, but i think its fine like this, no?
also, i want to remove the need of this complciated setup we have with `getVersionInfo` for registering a module, but havent investigated it yet in-depth.

kt3k:
> Can you provide some more information? What is this fixing?

This fixes the first text input box in https://deno.land/add_module, which currently always shows error for any module name. I think this prevents new authors publishing a new module

スクリーンショット 2022-08-12 23 30 04

kt3k (approved):
LGTM

lucacasonato (approved):
I am slightly concerned we'll have different rate limits on this endpoint, but I guess we'll see. Just be ready to revert if it goes sideways 😉

PR #2374, opened 2022-08-12
fix: use CDN for getVersionList

crowlKats:
Attempt 2 for #2372. only changes the url for `getVersionList` to use the CDN.


kt3k (approved):
LGTM

PR #2375, opened 2022-08-12
update hellobar

lambtron:


ry (approved):
LGTM 

I'm a little concerned about this being here 100% of the time. It makes the home page very busy.

PR #2403, opened 2022-08-23
feat: add module landing page

crowlKats:
- [x] display current version
- [x] mobile view
- [ ] properly handle relative references in readmes
- [ ] import block


kitsonk:
The display of the config is the same, irrespective of it being a `deno.json` or `deno.jsonc`.  Here is oak which has a `deno.jsonc`:

oak_v11_0_0___Deno

Either we should ensure it reflects the file, or we should change it do, `Contains Deno configuration` or something like that.

lucacasonato:
Why is having a Deno config a good metric? I don't want packages to add a Deno config file just because it looks good on the /x landing page.

kitsonk (approved):
LGTM - good enough to land IMO once CI passes.  A few niggles but we can iterate on it.

crowlKats:
> Why is having a Deno config a good metric? I don't want packages to add a Deno config file just because it looks good on the /x landing page.

@lucacasonato i wouldnt consider it a "good metric", its just an attribute

PR #2404, opened 2022-08-23
chore: use irm instead of iwr -useb

justjavac:


kt3k (approved):
LGTM. Thanks!

Note: We already use `irm` instead of `iwr -useb` in several other places:
- https://github.com/denoland/deno_install/issues/213
- https://github.com/denoland/manual/pull/385

PR #2417, opened 2022-08-27
update showcase image ultra

mashaal:
We have refreshed our website for an upcoming release, and wanted to update here.


ije (approved):
LGTM!

PR #2425, opened 2022-08-30
feat: publish search click events

kitsonk:
This add search event reporting when a user clicks on a search item in the global search box.

It generates a hashed version of the connection information presented to the server. This means there are no cookies dropped on a users machine, and the information is only used to attempt to uniquely identify users to algolia for its reporting and analytics.

### DIFF
--- a/components/Header.tsx
+++ b/components/Header.tsx
@@ -1,7 +1,6 @@
 // Copyright 2022 the Deno authors. All rights reserved. MIT license.
 
 /** @jsx h */
-/** @jsxFrag Fragment */
 import { h } from "preact";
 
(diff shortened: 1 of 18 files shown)


kitsonk:
Here is a view of the algolia event debugger indicating dispatched events are "correct" and starting to feed the engine:

Events___Algolia

components/Header.tsx
```diff
@@ -2,7 +2,7 @@
 
 /** @jsx h */
 /** @jsxFrag Fragment */
```
crowlKats, line 4:
```suggestion
```

crowlKats (approved):
LGTM

Issue #2364, opened 2022-08-10
Handle built-in APIs


kitsonk:
We should consider handling built in APIs on deno.land like we do on docland: https://doc.deno.land/deno/stable as well as we should publish the doc node to algolia to integrate into the search.

kitsonk:
Here is an example of what happens when searching for `WebSocket` which is common and clearly not the results we want to show:

<img width="1191" alt="Read_and_write_files___Manual___Deno" src="https://user-images.githubusercontent.com/1282577/183822103-be47bd73-f72e-4658-8c9f-cb1dfbe7a6d7.png">

dsherret:
Another example with `Deno.run` and `run`.
![image](https://user-images.githubusercontent.com/1609021/189983085-648605a6-171f-4198-8335-82eda5eed6ad.png)
![image](https://user-images.githubusercontent.com/1609021/189983119-1d54b28f-9749-47ea-b286-80f23afd87f5.png)

kitsonk:
This has landed.

PR #2428, opened 2022-08-30
feat: add built-in APIs

crowlKats:
Towards #2364

- [x] unstable toggle
- [x] landing page


kitsonk:
The index just feels messy at the moment.

Compare what we get for a index of a module:

std_0_155_0___Deno

Versus what we get for APIs:

API___Deno

Because the index of a module is broken up by symbol type, you get a nice alphabetical list for each section.  In built-ins the sorting by symbol type still happens in each section, making a very confusing mess when trying to read through it.  Feels like we should sort each section alphabetically.  I suspect it is also going to be a problem for the nav as well, in modules, you generally deal with a much smaller set of symbols, meaning the kind sort is slightly less problematic, though I have always felt it should be sorted alphabetically.

Also, there is really no context, indication for users unfamiliar with Deno what they are looking at.  Where authors have an opportunity to provide module doc for a module to provide context, we have nothing here.  Also the place where you look to figure out where you are (the grey bar at the top for std and x) is missing, feeling very inconsistent visually.

Maybe we can add some text, like for stable:

> There are APIs that are built into the Deno CLI that are beyond those that are built-ins for JavaScript. They are a combination of web platform APIs Deno has implemented and Deno specific APIs. We try to keep non-standard, Deno specific, APIs in the `Deno` namespace. We have grouped the APIs into the following functional categories.

For unstable:

> There are APIs that are built into the Deno CLI that are beyond those that are built-ins for JavaScript, including APIs that are unstable or experimental.  In order to use APIs marked as _unstable_, you will need to use `--unstable` on the command line to make them available. All the APIs are a combination of web platform APIs Deno has implemented and Deno specific APIs. We try to keep non-standard, Deno specific, APIs in the `Deno` namespace. We have grouped the APIs into the following functional categories.

It also feels like we should, in addition to the deprecated tag, display the unstable tag (and maybe the permission tags) in the index?

crowlKats:
> in addition to the deprecated tag, display the unstable tag (and maybe the permission tags) in the index?

Yes, unstable for sure; permission not sure. we need more space, will need a bit of a redesign

crowlKats:
> Also, there is really no context, indication for users unfamiliar with Deno what they are looking at. Where authors have an opportunity to provide module doc for a module to provide context, we have nothing here. Also the place where you look to figure out where you are (the grey bar at the top for std and x) is missing, feeling very inconsistent visually.

It does say `Runtime APIs` above the sidepanel, and has the `API` entry in the header highlighted, i think its fairly obvious where you currently are, and the gray bar at top for modules makes little sense to reuse for built-ins as we dont have breadcrumbs, popularity rating & description (i guess the latter could be could be worked around) for built-ins and wouldnt make sense. Also we are treating manual & built-ins as similar things for viewing and learning the functionalities of deno, and the gray bar wouldnt work for the manual.

kitsonk:
> and the gray bar at top for modules makes little sense to reuse for built-ins as we dont have breadcrumbs

namespace breadcrumbs? You totally lose your place here: https://dotland-9sj0b66nphxg.deno-staging.dev/api@v1.25.2?unstable=&s=Deno.errors.  It is arguable the title should be `Deno.errors` instead of just `errors` too.

> modules makes little sense to reuse for built-ins as we dont have breadcrumbs, popularity rating & description (i guess the latter could be could be worked around)

Though, the category could go up there where the description is, and the version list could be in the same place.  Outside of it being inconsistent, especially the index, it is just a "what the hell is this" problem... _we know, it is built-in APIs_ but we aren't the people that are going to be using it and it is too sparse and too unlike `std` and `x`.

kitsonk (approved):
LGTM

Ship it!

ship it

PR #2437, opened 2022-08-31
refactor: update module page style

ije:
a bit style update

### DIFF
--- a/routes/x/module.tsx
+++ b/routes/x/module.tsx
@@ -498,13 +498,15 @@ function VersionSelector({
   );
 }
 
-function InfoView(
-  { name, data, version }: {
-    name: string;
-    version: string;
-    data: ModInfoPage;
-  },
-) {
+function InfoView({
+  name,
+  version,
+  data,
+}: {
+  name: string;
+  version: string;
+  data: ModInfoPage;
+}) {
   data.description &&= emojify(data.description);
 
   const attributes = [];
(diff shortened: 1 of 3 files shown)


routes/x/module.tsx
```diff
@@ -375,87 +385,43 @@ function ModuleView({
...
-          data: data.data as DocPageSymbol | DocPageModule | DocPageIndex,
+          data: data.data,
           repositoryURL,
         }}
       />
     );
   }
-}
-
-function Breadcrumbs({
```
crowlKats, line 410: for what reason are the breadcrumbs removed? I think this is fairly important as else there is no way to navigate backwards; also, from recent comments, it seems people actively use the breadcrumbs, so i wouldnt be in favour of removing it

ije: you are tight, and i perfer to put it in the "nav" place instead of as a title, will update

ije: Screen Shot 2022-09-01 at 11 43 08
Screen Shot 2022-09-01 at 11 43 19

kitsonk: I like that, though the _Jump to..._ dramatically changes the payload that we deliver per page.  right now, we only deliver the nav based on the current folder, that would seem to imply we would need to deliver a whole module structure on every page render.  It is a non trivial amount of work as well as has performance implications.

Also, how would it look with something more deeply nested like: https://dotland-17mf6x94e9mg.deno-staging.dev/x/cliffy@v0.24.3/command/upgrade/provider/github.ts

ije: image

PR #2461, opened 2022-09-08
chore: update doc components and add replace

### LINKED ISSUE #2458: Parsing `$STD_VERSION` in markdown
A lot of the `std` JSDoc markdown includes `$STD_VERSION` which should be substituted with the version of `std` like the manual.

### LINKED ISSUE #2459: Inconsistent display of types
For example: https://deno.land/std@0.154.0/http/mod.ts?s=serve

std_0_154_0___Deno

We display the type for `handler` in the the signature, but not options.  Also, options should be displayed as optional in the signature.

crowlKats:
Fixes #2458
Fixes #2459


kitsonk (approved):
LGTM

PR #2466, opened 2022-09-08
fix: remove latest wording from modules

sigmaSd:
`Latest` is not accurate especially with deno that tends to cache responses, this can lead to misleading warnings see https://github.com/denoland/deno/issues/14933


sigmaSd:
The site is correct that the response is latest, but  because of deno caching this response will get outdated.

This is just a workaround idea, it might also be considered as unneeded hack

kitsonk:
It will be the latest at the time the error message is displayed in the CLI.  Thanks, but I prefer not to change it.

PR #2486, opened 2022-09-14
add cursor-pointer to menu and cross icon in header

mooxl:
I noticed that there is no pointer cursor in the mobile header element when hovering over the menu icon.


crowlKats (approved):
Thanks

PR #2487, opened 2022-09-14
add scroll behavior smooth to all pages

mooxl:
I know this change is debatable, but in my opinion, the smooth scroll improves the user experience by making it easier to follow jumps to anchors on the current page.

For comparison:

**with**

https://user-images.githubusercontent.com/35741000/190249548-eb4af88d-bc8e-405e-93a9-62037079a977.mp4

**without**

https://user-images.githubusercontent.com/35741000/190249559-2a291b1d-abe6-4720-b196-d4ae6b1ca9f1.mp4

### DIFF
--- a/components/Header.tsx
+++ b/components/Header.tsx
@@ -63,7 +63,7 @@ export function Header({ selected, main, manual }: {
 
             <label
               tabIndex={0}
-              class={tw`cursor-pointer lg:hidden ${
+              class={tw`lg:hidden ${
                 css({
                   "touch-action": "manipulation",
                 })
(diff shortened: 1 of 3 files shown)


crowlKats:
we actually already have this in the twind preflight, but seems it isnt working how it currently is, so instead of adding css to app.css, the preflight change should be adjusted

mooxl:
I don't know why, but `twind` doesn't know the rule `scroll-smooth` in the preflight. That's  why i had to switch to normal css.

crowlKats (approved):
LGTM, thanks

PR #2488, opened 2022-09-14
remove scroll into view on manual

### LINKED ISSUE #2485: Linking to a page header in the manual navigates to top of the page in Firefox browser
If I link to a specific header of a manual page in a new browser tab, for example https://deno.land/manual@v1.25.2/typescript/configuration#what-an-implied-tsconfigjson-looks-like, in Firefox it navigates to the top of the page instead.

Expected outcome (Google Chrome): 
afbeelding

Actual outcome (Firefox):
afbeelding

This does not happen with headers on the deno.land homepage, so it maybe only happens for the headers of the manual pages.

mooxl:
Since the redesign of the manual allowed us to collapse the navigation, we no longer need the scroll into view function of the links that I implemented back then.

Fixes #2485

### DIFF
--- a/routes/manual.tsx
+++ b/routes/manual.tsx
@@ -161,18 +161,7 @@ export default function Manual({ params, url, data }: PageProps<Data>) {
           </div>
         </div>
       </SidePanelPage>
-
       <Footer />
-
-      <script
-        dangerouslySetInnerHTML={{
-          __html: `
-        (function() {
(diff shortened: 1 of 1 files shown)


crowlKats:
Further below there is usage of `toc-active`, that should be deleted

crowlKats (approved):
LGTM, thanks

PR #2495, opened 2022-09-20
Use 'Official Deno project' instead of 'By Deno Team'

ry:


bartlomieju (approved):

routes/x/module.tsx
```diff
@@ -516,8 +516,8 @@ function InfoView(
     attributes.push(
       <div class={tw`flex items-center gap-1.5`}>
         <Icons.CheckmarkVerified class="h-4 w-auto" />
-        <span class={tw`text-tag-blue font-medium leading-none`}>
-          By Deno Team
+        <span class={tw`text-gray-600 font-medium leading-none`}>
```
crowlKats, line 519: why the color change? this doesnt fit with the design really, as the icons and the text have matching colors

ry: the other attribute has gray text

crowlKats: the other attribute has also a gray/black icon. another attribute, popularity tag, also has matching icon & color.

kt3k: I agree that `primary` color is confusing as it's not a link. How about changing `fill` property of svg's path to `currentColor` and style it via twind class?

crowlKats: Do you mean to then make the checkmark gray-600 as well?

kt3k: Yes, or maybe the same color as popularity tag (anything other than primary).

crowlKats: well popularity tag changes color depending on how popular so that wouldnt be great. i guess gray-600 might work

kt3k: Updated
スクリーンショット 2022-12-19 17 27 32

kt3k (approved):

kt3k:
Landing

PR #2496, opened 2022-09-21
feat: enrich page metadata

kitsonk:
This PR:

- Adds a dynamic component called `ContentMeta` which provides better management and information related to meta data associated with a page.  While it does not yet include images, it provides more meaningful meta information for cards and other expansions across the site.
- Alters the `isHTML` check to serve bots, like the Google Crawler and other bots and crawlers the HTML content for `std` and third party modules.  Currently, we serve the raw content, which is not what we really want indexed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,4 +35,4 @@ jobs:
           deno-version: v1.x
 
       - name: Run Tests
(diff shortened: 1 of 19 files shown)


crowlKats (approved):
LGTM

PR #2498, opened 2022-09-21
chore: minor fixes

lucacasonato:
- Use /api instead of doc.deno.land for CLI Runtime API.
- Fix dismissing search dialog with Escape


crowlKats (changes requested):
please stick with fully-qualified urls instead of relative; also this needs to be updated in doc_components as well, as we have the shared footer there (there is #2406 to use said shared footer)

crowlKats (approved):

PR #2500, opened 2022-09-21
feat: Add copy button to install commands

jespertheend:
I've made it only show on hover, but to be honest I think I like it better when they're always visible. Let me know what you think.

This uses the same method as the copy button for [module usage](https://github.com/denoland/doc_components/blob/ae83d2458df923a9a4b68113baebbbc1071a51ba/doc/usage.tsx#L100) which might be a bit hacky, but since `CodeBlock` is only used on the landing page I think it should be fine. Otherwise making this an actual island would be preferable I think.

https://user-images.githubusercontent.com/2737650/191625383-038d8887-e019-4e87-95d4-1469af5fe2c5.mp4


components/CodeBlock.tsx
```diff
@@ -132,6 +142,15 @@ export function RawCodeBlock({
           );
         })}
       </div>
+      {enableCopyButton &&
+      (
+        <button
+          className={tw`opacity-0 group-hover:opacity-100 rounded border border-[#D2D2DC] p-1.5 self-start`}
+          onClick={`navigator?.clipboard?.writeText('${code.trim()}');`}
```
jespertheend, line 149: I'm getting a type error here, but if I add a `// @ts-ignore` deno panics when running `deno fmt`, not sure if I can work around that.

crowlKats: thats known issue. dprint/dprint-plugin-typescript/issues/400. add a format ignore at the top of the file

routes/index.tsx
```diff
@@ -431,7 +432,12 @@ function InstallSection({ url }: { url: URL }) {
         </a>{" "}
         (Mac):
       </p>
-      <CodeBlock language="bash" code="brew install deno" url={url} />
+      <CodeBlock
+        language="bash"
+        code="brew install deno"
+        url={url}
+        enableCopyButton={true}
```
crowlKats, line 439: nitpick
```suggestion
        enableCopyButton
```

routes/index.tsx
```diff
@@ -441,6 +447,7 @@ function InstallSection({ url }: { url: URL }) {
         language="bash"
         code="irm https://deno.land/install.ps1 | iex"
         url={url}
+        enableCopyButton={true}
```
crowlKats, line 450: nitpick
```suggestion
        enableCopyButton
```

routes/index.tsx
```diff
@@ -420,6 +420,7 @@ function InstallSection({ url }: { url: URL }) {
         language="bash"
         code="curl -fsSL https://deno.land/install.sh | sh"
         url={url}
+        enableCopyButton={true}
```
crowlKats, line 423: nitpick
```suggestion
        enableCopyButton
```

routes/index.tsx
jespertheend: Ah I didn't even know this was possible, thanks!

crowlKats (approved):
LGTM, thanks

PR #2508, opened 2022-09-27
fix: keep symbol on implicit latest redirect

### LINKED ISSUE #2507: Redirecting to a symbol does not carry over the query param
for example [/std/testing/asserts.ts?s=assertEquals](https://deno.land/std/testing/asserts.ts?s=assertEquals) does not redirect to the symbol, the query params get dropped.

crowlKats:
Fixes #2507


kitsonk (approved):
LGTM

PR #2513, opened 2022-09-28
feat: improve social media images

kitsonk:
This changes the twitter card type to a large image format and also adds explicit twitter image tags, in addition it adds alt text to all the images.


hashrock (approved):
Great, thank you!
Checked with Slack and Twitter, rendered nicely!

PR #2515, opened 2022-09-29
feat: module symbol linking in docs

kitsonk:
This PR introduces module level symbol linking.  Check an example here: https://dotland-rgfb0y453my0.deno.dev/x/oak@v11.1.0/application.ts?s=Application where you can see many of the symbols referenced in the API are clickable.

The only thing I am not sure about is that the linked version of the symbol is _black_ versus _blue_, but that is currently what doc components renders.  @crowlKats thoughts?

### DIFF
--- a/util/doc_utils.ts
+++ b/util/doc_utils.ts
@@ -1,0 +1,24 @@
+// Copyright 2022 the Deno authors. All rights reserved. MIT license.
+
+import { DocPageNavItem } from "./registry_utils.ts";
+
+let currentSymbols: string[] | undefined;
+
+export function setNav(path: string, nav: DocPageNavItem[]) {
+  currentSymbols = undefined;
+  for (const item of nav) {
+    if (item.path === path && item.kind === "module") {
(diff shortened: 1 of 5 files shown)


util/doc_utils.ts
```diff
@@ -0,0 +1,54 @@
...
+): string | undefined {
+  if (!currentSymbols.size) {
+    return undefined;
+  }
+  if (namespace) {
+    const parts = namespace.split(".");
+    while (parts.length) {
+      const name = [...parts, symbol].join(".");
+      if (currentSymbols.has(name)) {
+        return `${current.pathname}?s=${name}`;
```
crowlKats, line 39: dont do this, but rather change the urlsearchparams obj of `current` & return `current.href`; this allows for this to work on /api?unstable

crowlKats (changes requested):
ragarding rendering color: its fine, the redering of symbol linking in docs needs to be cleaned up anyways; will look into it

crowlKats (approved):

PR #2517, opened 2022-10-03
fix: remove symbol from url if target url doesnt have it

crowlKats:
ie from symbol page, clicking in the sidepanel a file will 404 currently, this fixes it


kitsonk (approved):
LGTM

PR #2534, opened 2022-10-15
Add `Ctrl+K` to search bar

prettykool:
This is more of a thing for people who don't use/aren't familiar with the Mac keyboard layout. 
 
Here's what it looks like, now:
image


crowlKats:
Ideally this would only display the correct variant based on the system

crowlKats:
Closing as fixed by #2599

PR #2558, opened 2022-10-28
feat: add manual redirects

crowlKats:


ry (commented):
It would be good to have a test here...

kt3k:
@crowlKats Is this already safe to land in this state?

routes/manual.tsx
```diff
@@ -324,7 +302,40 @@ export const handler: Handlers<Data> = {
...
+          const slug = `${parentSlug}/${childSlug}`;
+          const name = typeof entry === "string" ? entry : entry.name;
+          tempList.push({
+            path: slug,
+            name,
+          });
+          if (typeof entry === "object" && entry.redirectFrom) {
+            for (const redirect of entry.redirectFrom) {
+              redirectList[redirect] = slug;
+            }
```
kt3k, line 320: We probably need to set correct `redirectFrom` arrays in toc.json of #400?

ry:
@crowlKats I've added a test stub for you, please add some tests.

kt3k:
I believe we need to create mapping from old pages to new pages and set them in `toc.json` in #400

Is this understanding correct?

for example `npm_specifiers` node becomes:
```
{
  "npm_specifiers": "Using Node.js Packages with npm Specifiers",
  ...
}
```
to:

```
{
  "npm_specifiers": {
    "name": "Using Node.js Packages with npm Specifiers",
    "redirectFrom": ["/node/compatibility_mode"]
  },
  ...
}
```

kt3k:
I wonder how many websites link to manual pages without version numbers..

kt3k (approved):
The change looks good to me. This adds the feature to set up redirects in TOC by `redirectFrom` property.

util/manual_utils_test.ts
```diff
@@ -41,3 +46,53 @@ Deno.test("get edit link new repo", () => {
...
+      "name": "Help",
+    },
+  };
+  const map = tocGen(toc, "foo");
+  assertEquals(
+    map,
+    new Map([
+      ["foo/basics", "Basics"],
+      ["foo/help", "Help"],
+    ]),
```
ry, line 96: It should test the redirect...

routes/manual.tsx
```diff
@@ -324,7 +287,18 @@ export const handler: Handlers<Data> = {
         }),
     ]);
 
-    return render!({ tableOfContents, content, version });
+    const { pageList, redirectList } = generateToC(
+      tableOfContents,
+      `/manual@${version}`,
+    );
+
+    const slashPath = "/" + params.path;
```
bartlomieju, line 295: Seems it can't handle a situation when `params.path` starts with a slash.

util/manual_utils_test.ts
```diff
@@ -41,3 +46,53 @@ Deno.test("get edit link new repo", () => {
...
+          },
+        },
+        "debugging_your_code": "Debugging Your Code",
+      },
+    },
+    "help": {
+      "name": "Help",
+    },
+  };
+  const { redirectList } = generateToC(toc, "foo");
```
bartlomieju, line 93: Please add some assertions for `pageList` as well

routes/manual.tsx
crowlKats: it never starts with a slash, so not a concern

util/manual_utils_test.ts
crowlKats: done

bartlomieju (approved):
LGTM

PR #2573, opened 2022-11-10
Improve the experience of the expand button in the manual

zhmushan:
`outline-none` is to remove the border when in focus.
image

`select-none` is to prevent the text on the right side from being selected when the user double-clicks the icon. PS: when the list is expanded and then clicked again to close it, it is considered a double-click.


ry (approved):
thanks

PR #2582, opened 2022-11-14
Fix the minus icon not appearing when expanding the expandable sections of the header

roj1512:


crowlKats (changes requested):
This breaks hover on desktop.

roj1512:
What hover do you mean?

roj1512:
@crowlKats What about this?

crowlKats:
You should be able to get this work by just moving the input to be the first item in the div and changing the children selectors; no need to add a new div.

roj1512:
I tried that but I cannot select the nth-child(2) with twind, can I?

crowlKats:
you could select the `even` child

crowlKats (approved):

PR #2584, opened 2022-11-14
decrese zindex for background

sigmaSd:
Currently the sub menu doesn't show up because its behind the background , this fixes it

[Screencast from 2022-11-14 20-53-33.webm](https://user-images.githubusercontent.com/22427111/201753290-8966251c-196c-4561-91b6-2e32d41a87fb.webm)

### DIFF
--- a/routes/index.tsx
+++ b/routes/index.tsx
@@ -42,7 +42,7 @@ export default function Home({ data }: PageProps<Data>) {
       <div>
         <Header />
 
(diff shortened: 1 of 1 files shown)


ry (approved):
Thanks!

sigmaSd:
You can increase they're individual zindex to fix it, but I'm not 100% confident if there is a better idea

PR #2589, opened 2022-11-14
various fixes

crowlKats:
- fix first icon section alignment
- better visibility and state for benchmark link
- add `"` around quote text


bartlomieju (approved):
LGTM

PR #2597, opened 2022-11-21
fix: dont send click event to toggle for desktop header

### LINKED ISSUE #2591: Goofy header behavior
If you click on a section on the header, it will keep that section open even if the user's mouse leaves the hover area. This results in some goofy looking behavior in specific circumstances.

Reproc steps:
1. Click "Modules" text
2. Hover over "Docs"

Fix:
- Either remove on-click behavior or disable on-hover behavior after the user has clicked on another nav element

Screenshot 2022-11-15 at 2 43 59 PM

crowlKats:
Closes #2591

### DIFF
--- a/components/Header.tsx
+++ b/components/Header.tsx
@@ -120,7 +120,7 @@ export function Header({ selected, manual }: {
                       <label
                         htmlFor={entry.content}
                         tabIndex={0}
-                        class="rounded-md flex items-center justify-between px-1 my-3 lg:(px-2 my-0)"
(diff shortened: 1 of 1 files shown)


WittySmirk:
### This change breaks mobile because you can no longer open the menus:
no_mobile

crowlKats:
@WittySmirk no it doesnt. this is only changed for the large breakpoint.

WittySmirk:
@crowlKats yeah mb I messed up the diff and had the pointer-events-none out of the lg scope.

kt3k (approved):
LGTM

Also confirmed that this fixes the issue with desktop width, and keeps toggle behavior on mobile width.

PR #2609, opened 2022-12-02
fix: Remove redundant icons

ry:
Note links in community drop down

Before
Screen Shot 2022-12-01 at 9 24 00 PM

After
Screen Shot 2022-12-01 at 9 23 43 PM


crowlKats:
@ry why? There is a reason why this was made as it is.

ry:
@crowlKats I don't think it's helping

PR #2618, opened 2022-12-08
feat: display dependencies

crowlKats:


kt3k:
`esm.sh` section of `fresh` looks strange to me. Is this a known issue?

スクリーンショット 2022-12-08 14 16 43

crowlKats:
yea the generated data from apiland is wrong; its something i will look into today

crowlKats:
> How about placing `Dependencies` section above `Versions` section? I think dependencies are more interesting info to users than past versions.

hm I am not sure. @hashrock what do you think?

hashrock:
I agree! let's try

kt3k:
FYI npm registry shows the tabs in the order of `README` -> `Code` -> `Depenedencies` -> `Dependents` -> `Versions`

crowlKats:
PTAL

routes/x/module.tsx
```diff
@@ -728,6 +741,35 @@ function InfoView(
...
+              <div class="text-gray-400  text-sm leading-4">
+                Dependencies
+              </div>
+              {Object.entries(dependencies).sort(([kindA], [kindB]) =>
+                kindA.localeCompare(kindB)
+              ).map(([kind, dependencies]) => (
+                <div>
+                  <>
+                    <div class="font-medium">{kind}</div>
+                    <div class="children:(inline-block max-w-full mr-2 link truncate)">
```
kt3k, line 755: スクリーンショット 2022-12-08 14 31 09

When I first see this, I missed `ts_morph@16.0.0` existing (it looked like a part of version name of `rutt@..`. How about making this `children:block` instead of `children:inline-block` and list all deps vertically?

routes/x/module.tsx
```diff
@@ -777,6 +819,110 @@ function InfoView(
   );
 }
 
+function generateURL(
```
kt3k, line 822: Nice util! But let's move this to `util/` and add some test cases for each registry entry as it's difficult to test each switch branch in UI

crowlKats: yep, will do. was actually thinking of moving this to apiland

kt3k (approved):
LGTM. Great feature!

PR #2628, opened 2022-12-19
chore: update std

crowlKats:


bartlomieju (approved):
LGTM

import_map.json
```diff
@@ -1,7 +1,7 @@
 {
   "imports": {
     "@/": "./",
-    "https://deno.land/std@0.154.0/": "https://deno.land/std@0.153.0/",
+    "https://deno.land/std@0.154.0/": "https://deno.land/std@0.170.0/",
```
bartlomieju, line 4: Is this used in one of the dependencies?

crowlKats: this was a problem with esm.sh a while ago. dont want to break things for now, so this should be fine. will look if updating the pins on those esm deps will be fine later

Issue #2632, opened 2022-12-23
Module name doesn't nicely wrap if it's longer than the column width on deno_x


itsdouges:
<img width="360" alt="image" src="https://user-images.githubusercontent.com/6801309/209263611-5a2cca14-aa5b-422f-a806-13dd73758ce6.png">

PR #2633, opened 2022-12-23
fix: prevent wrapping and truncate overflowing text

yakuppgeyikk96:
#2632 Removal of "children:inline-block" class which causes to wrap


crowlKats (changes requested):
I am not in favour of this fix, as we explicitly added this to deal with too long pahts when viewing documentation.

Issue #2634, opened 2022-12-23
remove symbol search


crowlKats:
- [x] remove from dotland (#2635)
- [x] remove from apiland (https://github.com/denoland/apiland/pull/76)
- [x] delete entries in algolia

PR #2635, opened 2022-12-23
remove symbol search

### LINKED ISSUE #2327: Symbol search results should prioritise default modules
Currently searching for `assert` returns quite a few clearly "internal" modules of a repo:

Third_Party_Modules___Deno

If a module has a default module, we should either not publish the symbols of the other modules inside that module _or_ ensure they are a much lower priority after other modules indexes.

In the above example, we would want to see `std` and `conditional_type_checks` before the other modules.

### LINKED ISSUE #2362: "serve" as default search term skews results
We are currently defaulting to "serve" as a search result, meaning it gets a lot of hits, skewing the results.  We need to figure a better way to manage this.

### LINKED ISSUE #2516: Global search shouldn't be global when in a module
If you are in std or a third party module and you open the search, you would more likely expect:

- it to default to symbol search
- default to only searching within the current module

The first one should be straight forward.  The second one should be possible, though might need some re-work of things.

cc/ @surma

crowlKats:
Towards #2634

Closes #2327
Closes #2362
Closes #2516

### DIFF
--- a/components/Header.tsx
+++ b/components/Header.tsx
@@ -5,7 +5,6 @@ import { css } from "twind/css";
 import type { JSX } from "preact";
 import * as Icons from "./Icons.tsx";
 import GlobalSearch from "@/islands/GlobalSearch.tsx";
-import versions from "@/versions.json" assert { type: "json" };
 
 interface HrefEntry {
   content: string;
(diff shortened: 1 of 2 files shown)


ry (approved):
LGTM - I'm interested to hear how much people complain

sigmaSd:
This was very useful, why was it removed ?

KnorpelSenf:
> I'm interested to hear how much people complain

I'd also like to complain, so now there are 2 people who found this useful :D

sigmaSd:
I feel like people didn't know that this even exists, since it was not that much advertised and was not up for long. The problem now is even std symbols are not searchable anymore

PR #2637, opened 2022-12-24
feat: add `$MODULE_VERSION` markdown replacer

### LINKED ISSUE #2494: Variables in the README.md if a module
### Discussed in https://github.com/denoland/deno/discussions/15961

<div type='discussions-op-text'>

<sup>Originally posted by **justinmchase** September 20, 2022</sup>
In most / all readme's there is usually a little usage example to get someone started. It will usually start with a line that shows how to import the current module. That import statement usually should have a version number in it. Trying to keep that version number up to date with the actual latest version tag isn't super hard but it does require a little automation, or dilligence, or both.

What would be nice is if deno.land/x recognized, at least, the $GIT_TAG variable and just did a straight replacement, so if I had some snippet in my readme such as: like:

````ts
```ts
import { example } from "https://deno.land/x/example@$GIT_TAG/mod.ts";
```
````

Then it would render like this, when viewed relative to the tag named `0.1.0` in the deno land third party module UI.
```ts
import { example } from "https://deno.land/x/example@0.1.0/mod.ts";
```</div>

crowlKats:
Closes #2494

### DIFF
--- a/components/DocView.tsx
+++ b/components/DocView.tsx
@@ -21,9 +21,11 @@ export function DocView({
 
   data,
 }: CommonProps<DocPageSymbol | DocPageModule | DocPageIndex>) {
-  const replacer: [string, string][] | undefined = name === "std"
+  const replacer: [string, string][] = name === "std"
     ? [["$STD_VERSION", version]]
-    : undefined;
+    : [];
+  replacer.push(["$GIT_TAG", version]);
+
   const baseUrl = new URL(url);
   baseUrl.pathname = getModulePath(name, version);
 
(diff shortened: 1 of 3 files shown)


kt3k:
How about `$MODULE_VERSION` instead of `$GIT_TAG`?

crowlKats:
sure

kt3k (approved):
LGTM

I'd request another approval from core team member as hard-coding this affects the entire registry

bartlomieju (approved):
LGTM

PR #2686, opened 2023-02-13
fix: Give better feedback validating the module name

bastilian:
This adds a distinction between a valid module name and a module name that has already been taken. It also clarifies what a valid module name is.

This addresses the confusion pointed in https://github.com/denoland/dotland/issues/2393.

https://user-images.githubusercontent.com/7757/218585985-e998b151-7f96-4046-a6b3-62bcc0a3aae1.mp4

### DIFF
--- a/utils/registry_utils.ts
+++ b/utils/registry_utils.ts
@@ -3,6 +3,7 @@
 import type { DocNode } from "deno_doc/types";
 import type { LibDocPage, ModuleEntry } from "$apiland_types";
 
+const NAME_REGEX = /^[a-z0-9_]{3,40}$/;
 export const CDN_ENDPOINT = "https://cdn.deno.land/";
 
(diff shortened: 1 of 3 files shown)


utils/registry_utils.ts
```diff
@@ -358,3 +359,17 @@ export function getDocAsDescription(
...
+  name: string,
+  controller: AbortController,
+) {
+  if (name === "" || !NAME_REGEX.test(name)) {
+    return "invalid";
+  } else {
+    return await getVersionList(name, controller.signal)
+      .then((e) => !e)
+      .catch(() => false)
+      .then((e) => e);
```
crowlKats, line 373:
```suggestion
      .catch(() => false);
```

crowlKats (approved):
LGTM

PR #2689, opened 2023-02-17
feat: add useful com redirects

jordanreger:
This is a simple change but would allow erroneous links (like https://deno.land/deploy instead of https://deno.com/deploy) to still convert correctly. These are the three paths that I think are probably often mistaken.


crowlKats:
closing as this has been addressed

PR #2693, opened 2023-02-21
fix: adjust `isOldVersion` to match old behaviour

crowlKats:


kt3k (approved):
LGTM. What is the possible invalid version input?

crowlKats:
`main` of a git hash (used for manual)

PR #2698, opened 2023-02-28
fix: hardcode default global search results

crowlKats:


islands/GlobalSearch.tsx
```diff
@@ -78,13 +78,90 @@ const client = algoliasearch("QFPCRZC6WX", "2ed789b2981acd210267b27f03ab47da", {
   requester,
 });
 
+const defaultResult: Results = {
```
kt3k, line 81: Let's comment how these data are created and can be updated in the future above this line

kt3k:
Note to the motivation: This replaces the search result for the empty input (which is queried by default when the search panel is opened) with the hard-coded data, and this saves the cost of algolia usage

kt3k (approved):
LGTM

PR #2703, opened 2023-03-06
showcase: Add socialcards.dev app

MariusVatasoiu:
I'd like to add [socialcards.dev](https://socialcards.dev) to the Deno showcase.

It's a full stack application built with Deno, using modules like oak, denoauth, oak_sessions, mongo and many more.


kt3k (approved):
LGTM

PR #2705, opened 2023-03-08
Fix the Netlify logo size

lucacasonato:
The text was comparatively way too small. It's now scaled to be the same
size as other logos.


hashrock (approved):
Nice catch!

PR #2710, opened 2023-03-12
chore: add info on viewing benchmarks for a specific number of commits

GJZwiers:


routes/benchmarks.tsx
```diff
@@ -186,7 +186,8 @@ export default function Benchmarks({ url, data }: PageProps<Data>) {
               >
                 {!showAll ? "all" : "the most recent"}
               </a>{" "}
-              commits.
+              commits or append ?-n to this page's URL to view results for the
```
bartlomieju, line 189: Could you give an example here so it could be copy-pasted? `?-100` would work

GJZwiers: Done!

bartlomieju (approved):
Thanks

PR #2718, opened 2023-03-17
showcase, var - CLI tool for variables

denyzhirkov:
here is the source - https://github.com/denyzhirkov/var


denyzhirkov:
Something strange with tests and cheks

PR #2740, opened 2023-04-03
add logs

crowlKats:


bartlomieju (approved):
LGTM, as long as we can handle additional overhead

PR #2742, opened 2023-04-05
Update web compatibility images

hashrock:
- recreate assets with figma
- use svg to prevent blurry images


crowlKats (approved):
LGTM, I like the switch to SVG!

PR #2755, opened 2023-05-10
Use marked-smartypants to replace punctuation in markdown content with smart punctuation

roj1512:
Waiting on https://github.com/denoland/deno-gfm/pull/66.

### DIFF
--- a/components/Markdown.tsx
+++ b/components/Markdown.tsx
@@ -1,6 +1,11 @@
 // Copyright 2022-2023 the Deno authors. All rights reserved. MIT license.
 
-import { render } from "$gfm";
+import { Marked, render } from "$gfm";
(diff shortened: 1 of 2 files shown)


kt3k:
What problem does this solve?

roj1512:
https://github.com/denoland/manual/issues/643

kt3k:
That seems fixed with https://github.com/denoland/manual/pull/644

roj1512:
Yes, we made the quotes consistent in denoland/manual so we can proceed with this. I’ve also explained everything to @lambtron:

We should be using normal punctuation (e.g. ""'') in the markdown source, but convert them to automatically smart punctuation.

lambtron (approved):

